PR #44, opened 2025-06-04
Add initial MCP support

Steve-Mcl:
## Description

- Adds MCP Client and Server with one `promptTemplate` named `explain_flows`
- Adds nls message catalog for better prompt action naming
    - image

### NOTES:
- At this early time, this implementation uses an internal MCP client and server, with the only connection made being `inMemoryTransport`.  That means the MCP server cannot be _used_ by other agents (e.g. by STDIO or HTTPStreaming/SSE)

### TODO
- [x] Sanitise the nodes array (identify and remove unnecessary keys to reduce token usage)

### For Consideration (this or future iteration)
- [ ] More granular error/response handling (e.g. let user know if it is "too large", if it is an "upgrade" feature, etc.)
- [ ] Review and improve UX (currently, the action is available only as an action list entry and the output is a simple notification)
- [x] Figure out why providing a label for the actions is not used (label is currently a sanitisation of the action name)
    - seems to be that `options.label` is ONLY a look up for i18n.
    - plugin  does not load 18n catalog (tried index.json, index.json, index.json and several others), `/plugins/messages?lng=en-US` is never populated!
- [ ] Architecture review:
    - Currently the MCP Client and Server are only communicable via inMemoryTransport
- [ ] Consider Including the parent flow name (as context cue)
- [ ] Consider augmentation by allowing the user to add context e.g to specify things like "I am only interested in the output of the last node"

### Other points to think on
* Gating functionality based on team type? (TBD)
* Safeguards:
    - As the MCP promptTemplate is generated in the instance and its compiled prompt passed as-is to the AI, there is effectively no validation at the final stage.
   - If we decide to revoke the feature, we cannot prevent it appearing in the UI, as it is a custom action.
        - For this, I had started working on a distributed architecture - where MCP promptTemplates are generated at the AI instance and each Node-RED pulls them locally, but this is not yet finalised (and poses difficulties due to executable/non-serialise-ability of the promptTemplates).
 

## Related Issue(s)

https://github.com/FlowFuse/node-red/issues/33

## Checklist

 - [ ] Suitable unit/system level tests have been added and they pass 
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/index.html
+++ b/index.html
@@ -24,6 +24,7 @@
             requestTimeout: AI_TIMEOUT
         }
         let assistantInitialised = false
+        let assistantMCPOneTimeFlag = false
         debug('loading...')
         RED.plugins.registerPlugin('flowfuse-nr-assistant', {
             type: 'assistant',
@@ -36,6 +37,13 @@
                         assistantOptions.enabled = !!msg?.enabled
                         assistantOptions.requestTimeout = msg?.requestTimeout || AI_TIMEOUT
                         initAssistant(msg)
+                    }
+                    if (topic === 'nr-assistant/mcp/ready' && !assistantMCPOneTimeFlag) {
+                        assistantMCPOneTimeFlag = true
+                        if (assistantOptions.enabled) {
+                            debug('assistant MCP initialised')
+                            RED.actions.add('flowfuse-nr-assistant:explain-selected-nodes', explainSelectedNodes, { label: '@flowfuse/nr-assistant/flowfuse-nr-assistant:actions.explain-selected-nodes' })
+                        }
                     }
                 })
             }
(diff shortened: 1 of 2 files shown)


Steve-Mcl:
Review tip:

`index.html` [lines 57 ~ 347](https://github.com/FlowFuse/nr-assistant/pull/44/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051L57-L347) can be pretty much ignored. This was an indentation fix to remove a useless `} else { ... }` wrapping - the GH diff is not figuring this out!

The changes to that file begin at  [line 350](https://github.com/FlowFuse/nr-assistant/pull/44/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051L57-L347)

Steve-Mcl:
@joepavitt if you want to pull and test, let me know and i'll get you up and running.

joepavitt:
Working great - follow up actions to be detailed in GH issues, but nothing blocking

Steve-Mcl:
Last minute check on legacy device agent (node 14, agent 2.x) revealed we need to handle the fact node14 cannot load MCP libs.
[5d5f8fc](https://github.com/FlowFuse/nr-assistant/pull/44/commits/5d5f8fc595def7eba08cd65eeaab080ddeaaaded) includes a patch to prevent addition of action should MCP fail to initialise.

PR #58, opened 2025-06-26
Implement node suggestions

Steve-Mcl:
### Foreword
This is a large PR. The major portions are
1. Refactoring of most logic in `index.js` to a common shared lib `assistant.js`
2. Addition of a 2nd plugin `completions.html`
3. Addition of tensorflow, asset fetching and completions logic

Local test + a quick scan of the code (for obvious issues) is recommended.
Extensive Unit tests were added to provide a level of confidence.

### UPDATE:
Following Eng meeting and change of how assets will be loaded, this has been refactored to only request the assets upon frontend editor signalling to the backend that they are required. This will hugely reduce the chance of any bursts when lage pipelines are triggered.

### TLDR;
* NR 4.1 introduces a new API for setting suggestions and requires a new type of plugin `type: 'node-red-flow-suggestion-source'` - `completions.html` introduces this
   * Because we now have 2 plugins in this package, backend logic from `index.js` was refactored into a new shared file `lib/assistant.js` for both the original plugin `index.html` and new plugin `completions.html` to share
* To provide completions, there was a requirement to load TensorFlow - the package is updated for the inclusion of this
* All original FE logic remains in `index.html`

#### 1. Frontend Initialisation (`completions.html`, `index.html`)

**Event-Driven Loading:**  
The frontend listens for backend events via `RED.comms.subscribe('nr-assistant/#', ...)`:
- `'nr-assistant/initialise'`: Sets up options and enables features.
- `'nr-assistant/mcp/ready'`: Indicates the Model Context Protocol (MCP) is ready; triggers loading of completions if supported.
- `'nr-assistant/completions/ready'`: Indicates completions (e.g., TensorFlow model) are loaded and ready.

NOTE: Features are gated behind checks for `assistantOptions.enabled`, `mcpReady`, and `completionsReady`.

---

#### 2. Backend Initialisation (`lib/assistant.js`)

**Singleton Pattern:**  
The core backend logic is not located in singleton class instance `Assistant`. It provides the httpAdmin endpoints for the frontend, loads libraries (MCP/TF/etc) and handles proxying calls to the assistant API upstream

**`init` Method:**  
- Sets up internal state and checks for required configuration (`enabled`, `url`, `token`).
- Publishes `'nr-assistant/initialise'` to the frontend with client settings.
- Loads MCP (Model Context Protocol) if enabled and supported.
- Loads completions (TensorFlow.js model and vocabulary) if enabled and supported, **after** MCP is ready.
- Registers admin HTTP endpoints for API calls.

**MCP and Completions Loading:**  
- MCP is loaded via dynamic import and sets `mcpReady` on success.
- Completions are loaded only after MCP is ready, using dynamic imports for TensorFlow.js and model/vocabulary files.
- Backend publishes `'nr-assistant/mcp/ready'` and `'nr-assistant/completions/ready'` events to the frontend.

**Degraded Mode:**  
If MCP or completions fail to load, the plugin logs warnings and operates in a reduced functionality mode.

---

#### 3. Frontend/Backend Coordination

**Retained Events:**  
Backend uses `RED.comms.publish(..., true /* retain */)` to ensure the frontend receives the latest state on connect/reload.

**Feature Flags:**  
Both frontend and backend check for feature enablement before attempting to load or use advanced features.

**Loading Sequence:**
1. Backend initialises and publishes `'nr-assistant/initialise'`.
2. Backend loads MCP, then publishes `'nr-assistant/mcp/ready'`.
4. When the frontend is loaded for the first time, it will send `'nr-assistant/completions/load'`, instructing the backend to fetch assets for completions
5. Backend loads completions (if enabled), then publishes `'nr-assistant/completions/ready'`.
6. Frontend listens for these events and enables UI/features accordingly.

---

#### 4. Error Handling

**Frontend:**  
If features are not enabled or ready, UI actions are disabled/hidden accrdingly

**Backend:**  
Returns HTTP 503 if not ready, logs errors, and sends error responses for failed initialisation or requests.

### New unit tests added for the refactored / shared components
```
assistant
    ✔ should be constructed
    ✔ should initialize with valid settings (153ms)
    ✔ should not re-initialize if already initialized
    ✔ should not be enabled if disabled in settings
    ✔ should not be enabled if required url option is missing
    ✔ should not be enabled if required token option is missing
    ✔ should skip loading completions for node-red < 4.1
    ✔ should continue to finish loading but with degraded functionality if MCP fails to load
    ✔ should continue to finish loading but with degraded functionality if the model is unavailable (115ms)

  flowGraph
    getLongestUpstreamPath()
      ✔ should return the longest upstream path for a flow
      ✔ should return an empty array if there are no upstream connections
      ✔ should handle multiple branches and pick the longest path
      ✔ should throw an error on circular reference
      ✔ should throw an error on circular reference to self
      ✔ should return empty array if node does not exist
      ✔ should handle disconnected nodes

  utils
    tryJsonParse
      ✔ should parse valid JSON string
      ✔ should return defaultValue for invalid JSON
      ✔ should return undefined for invalid JSON if no defaultValue is provided
      ✔ should parse JSON arrays
      ✔ should parse JSON null
      ✔ should not throw on bad JSON
    isObject
      ✔ should return true for plain objects
      ✔ should return false for arrays
      ✔ should return false for null
      ✔ should return false for strings
      ✔ should return false for numbers
      ✔ should return false for functions
    hasProperty
      ✔ should return true if object has property
      ✔ should return false if object does not have property
      ✔ should return false if not an object
      ✔ should return false for inherited properties
      ✔ should return true for own property even if value is undefined

  sharedUtils
    cleanFlows
      single node
        ✔ should clean a single node and remove internal properties
        ✔ should return null for null or invalid input
        ✔ should return null if node is not cleaned (no id)
      flow array
        ✔ should remove internal properties from nodes
        ✔ should clean nested group nodes recursively and return the count all nodes
        ✔ should avoid circular references using visited set
        ✔ should not exceed MAX_DEPTH
        ✔ should handle empty input array
        ✔ should filter out null nodes
```

## Checklist

 - [x] Suitable unit/system level tests have been added and they pass 
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl:
This has been superseded by #62

Steve-Mcl:
> @Steve-Mcl let me know once this is good to go (assuming we're switching out TF for ONNX now?) I'll mark it as a draft for now

I opted to branch and make a new PR - it is now ready #62

Steve-Mcl:
PS, please dont delete this branch for now

PR #89, opened 2025-11-28
Allow the assistant to be installed in standalone Node-RED instances

knolleary:
## Description

This PR enables the assistant to be installed into any Node-RED instance outside of FlowFuse.

When run in this way, it enables 'standalone' mode. In this mode it will show a login menu option.

image

This will direct your to FlowFuse Cloud to login and authorise access to your account. Note, this is the generic OAuth application approval screen. It could be futher customised for FF Assistant, but that's on the platform side so can be iterated separately.

image

Once enabled, the UX is the same as in hosted instances:

image

The only FF Assistant feature that is *not* enabled is the Tables content assist - as the Tables feature is FF-only so doesn't make sense in standalone mode.

---

This has required quite a lot of reworking of the lifecycle of the plugin. Previously it simply had to check once on startup whether proper configuration was provided and either enable the AI features or do nothing more.

Now the features can be enabled (and disabled!) asynchronously to startup. To be clear - I've done what feels like the bare minimum to get this working; if the plugin was to be written from scratch with this mode a requirement from the start, I suspect it'd have a different structure.

The code uses the presence of the ff settings block in settings.js to determine whether its in standalone mode or not.

I have tested various lifecycle transitions around login, token refresh, token expiry etc and think its working as desired. 

---

The server side changes needed to support this have already been merged to FF Cloud - I have tested against it successfully.

### DIFF
--- a/test/unit/lib/assistant.test.js
+++ b/test/unit/lib/assistant.test.js
@@ -231,9 +231,9 @@ describe('assistant', () => {
         assistant._mcpClient.should.be.an.Object()
         assistant._mcpServer.should.be.an.Object()
 
-        RED.log.info.calledWith('FlowFuse Assistant Model Context Protocol (MCP) loaded').should.be.true()
-        RED.log.info.calledWith('FlowFuse Assistant Completions Loaded').should.be.true()
-        RED.log.info.calledWith('FlowFuse Assistant Plugin loaded').should.be.true()
+        RED.log.info.calledWith('FlowFuse Expert Model Context Protocol (MCP) loaded').should.be.true()
+        RED.log.info.calledWith('FlowFuse Expert Completions Loaded').should.be.true()
+        RED.log.info.calledWith('FlowFuse Expert Plugin loaded').should.be.true()
 
         RED.log.error.called.should.be.false()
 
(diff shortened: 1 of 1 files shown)


lib/auth/index.js
```diff
@@ -0,0 +1,277 @@
...
+}
+
+/**
+ * Make an API request to refresh the user token
+ * @param {*} token
+ */
+async function refreshToken (token) {
+    const params = {
+        grant_type: 'refresh_token',
+        client_id: 'ff-plugin',
```
Steve-Mcl: Any benefit in setting a unique (and identifiable) client ID like `ff-plugin:assistant` ?

lib/assistant.js
```diff
@@ -410,7 +441,7 @@ class Assistant {
 
     // #region Admin Endpoints & HTTP Handlers
 
-    initAdminEndpoints (RED, { inlineCompletionsEnabled }) {
+    initAdminEndpoints (RED) {
```
Steve-Mcl: `inlineCompletionsEnabled` param has been removed from the signature but it is still passed in the call at line 177.

index.html
```diff
@@ -1435,7 +1517,7 @@
                                 options.buttons.unshift(
                                     {
                                         text: plugin._('explain-flows.dialog-result.copy-button'),
-                                        icon: 'ui-icon-copy',
+                                        // icon: 'ui-icon-copy',
```
Steve-Mcl: Is removing button icon a deliberate choice?

index.html
```diff
@@ -1452,7 +1534,7 @@
                                 options.buttons.unshift(
                                     {
                                         text: plugin._('explain-flows.dialog-result.comment-node-button'),
-                                        icon: 'ui-icon-comment',
+                                        // icon: 'ui-icon-comment',
```
Steve-Mcl: Is removing button icon a deliberate choice?

index.html
```diff
@@ -1421,7 +1503,7 @@
                         const options = {
                             buttons: [{
                                 text: plugin._('explain-flows.dialog-result.close-button'),
-                                icon: 'ui-icon-close',
+                                // icon: 'ui-icon-close',
```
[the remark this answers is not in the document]
knolleary: Yes - we don't have icons on any other dialog button in Node-RED so always felt out of place here. Took the opportunity to get rid.

Issue #103, opened 2026-01-08
Support actions from FF Expert


Steve-Mcl:
### Description

To facilitate operations commanded from the FF Expert, a transport between the parent and the iframe that the immersive NR lives in is required. 

This transport should only operate in a FF env and only when the instance is in a FF immersive env.

See parent issue for more details.

PR #106, opened 2026-01-14
Add support expert actions

Steve-Mcl:
### Features implemented in this PR:

1. Establish postMessage communication between the NR iframe (via nr-assistant) to the ff-expert in the parent window.
2. Send init message from ff-expert to nr-assistant when the iframe is loaded.
3. Add support for requesting assistant version and supported actions (self describing)
4. Implement "Import Flow" feature to permit flows to be imported into NR from the ff-expert.
5. Implement "Install Node" feature to permit nodes to be installed into NR from the ff-expert.
6. Send editor open/close events from nr-assistant to ff-expert (see notes below).

#### Flow Import feature

In the NR editor, `RED.clipboard.import` is the function used to import (and if necessary) show import
the conflicts dialog but, because `RED.clipboard.import` does not accept a "flow" parameter I had to
duplicate the internal NR logic and call `RED.view.importNodes` in the plugin with `generateIds:true`.

Ideally, the inner function `importNodes` (in `clipboard.js`) would be the exposed (so the internal logic
for showing conflicts would be called)
OR
the exposed import function  `RED.clipboard.import` should include a `flow` parameter to allow importing
provided flows (and probably `dontShowImportDialog [default=false]` and `performConflictResolution [default=true]` options).

Short of updating the NR core to do this, to enable the same conflict resolution in the nr-assistant "import flows"
capability, I would have to re-implement the entire conflict resolution dialog and logic in the plugin which is not ideal.

Lastly, if a shade is open (i.e. a node edit dialog is open) when the import is operated, the flow is actually imported
and dropped onto the workspace behind the shade. This is not ideal (especially on a small screen where the edit dialog
and the expert cover most of the workspace). It would be better if NR core handled this scenario by either
1) preventing the import when a shade is open
or
2) automatically closing the shade, performing the import, then reopening the shade.

##### Recommendation / follow up
- [x] Raise an issue in the NR core repo to expose the inner `importNodes` function or add parameters to `RED.clipboard.import`. 
   - https://github.com/node-red/node-red/issues/5432
- [x] Raise an issue in the FlowFuse repo to a) auto close the expert when an import is initiated from the expert, b) prevent import when a shade is present.
   - https://github.com/FlowFuse/flowfuse/issues/6512

#### Install Node feature

In its current form, the "Install Node" feature in the nr-assistant will open and populate the "Manage palette" dialog in NR core.
However, if there is a shade (i.e. editor is open), the dialog will open but due to some weird timing issue with NR core, the dialog
the wrong tab is be shown and the search field is not populated with the node to install.

This could be mitigated by only permitting the FF-Expert operation when there is no shade (i.e. no node edit dialog open) however
it would be better if the NR core "Manage palette" dialog logic was updated to handle this scenario correctly.

For now, we have added events `editor:open` and `editor:close` to the plugins post messages so that the ff-expert can disable
the "Install Node" feature is an editor is open (the code in the ff-expert to enable/disable the feature is not yet implemented).

Additionally, if the user proceeds with the install operation when, when the toast notification is shown in NR core, it is eclipsed
by the ff-expert window. The user has to manually close the ff-expert window to see the notification on small screens. This could be mitigated
by auto closing the ff-expert window when the install operation is initiated.

##### Recommendation / follow up
- [x] Raise an issue in the NR core repo to fix the "Manage palette" dialog so that it works correctly every time.
   - https://github.com/node-red/node-red/issues/5433
- [x] Raise an issue in the FlowFuse repo to a) auto close the expert when an install node is initiated from the expert, b) disable the install node feature when a shade is present.
   - https://github.com/FlowFuse/flowfuse/issues/6512

## Related Issue(s)

#103

## Checklist

 - [x] Suitable unit/system level tests have been added and they pass 
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Link to Changelog Entry PR, or note why one is not needed.

## Labels

### DIFF
--- a/test/unit/lib/assistant.test.js
+++ b/test/unit/lib/assistant.test.js
@@ -96,6 +96,7 @@ describe('assistant', () => {
     let assistant
     /** @type {import('got').Got} */
     let fakeGot
+    const packageVersion = require('../../../package.json').version
 
     beforeEach(() => {
         // first, delete the cached assistant module
@@ -209,6 +210,7 @@ describe('assistant', () => {
         RED.comms.publish.calledThrice.should.be.true()
         RED.comms.publish.firstCall.args[0].should.equal('nr-assistant/initialise')
         RED.comms.publish.firstCall.args[1].should.eql({
+            assistantVersion: packageVersion, // required for the frontend to support ff-expert features
             enabled: true,
             tablesEnabled: false,
             requestTimeout: 60000,
@@ -217,6 +219,7 @@ describe('assistant', () => {
 
         RED.comms.publish.secondCall.args[0].should.equal('nr-assistant/mcp/ready')
         RED.comms.publish.secondCall.args[1].should.eql({
+            assistantVersion: packageVersion, // required for the frontend to support ff-expert features
             enabled: true,
             tablesEnabled: false,
             inlineCompletionsEnabled: false,
@@ -251,6 +254,7 @@ describe('assistant', () => {
         RED.comms.publish.called.should.be.true()
         RED.comms.publish.firstCall.args[0].should.equal('nr-assistant/initialise')
         RED.comms.publish.firstCall.args[1].should.eql({
+            assistantVersion: packageVersion,
             enabled: true,
             tablesEnabled: false,
             inlineCompletionsEnabled: true,


cstns (approved):
LGTM

PR #235, opened 2026-04-02
feat: add automation/add-nodes action

### LINKED ISSUE #196: Support for Add Nodes action
### Support for Add Nodes action
**Action:** `automation/add-nodes`

Add support for the `automation/add-nodes` action in `ExpertAutomations`.

This action receives an array of raw node JSON objects and adds them to the live NR4 canvas without restarting the runtime.

Raw JSON is not directly usable with `RED.nodes.add()` — it skips setting `inputs`/`outputs`, `_config`, and `_def`, which causes missing port handles and the orange misconfigured triangle. The implementation must mirror what `RED.nodes.importNodes()` does internally:
- Look up the type definition via `RED.nodes.getType(node.type)` and set `_def`, `inputs`, `outputs`
- Populate `_config` by iterating `def.defaults` so `validateNode()` passes
- Call `RED.nodes.add(node)` then `RED.editor.validateNode(node)`
- After all nodes are added: switch to the target tab if different from the active one (fires `workspace:change` automatically), otherwise emit `workspace:change` manually to rebuild the `activeNodes` list that D3 binds to

**Params:** `{ nodes: NodeJSON[] }` — each node must include `id`, `type`, `z` (tab ID), `x`, `y`, and any type-specific properties.

### Epic/Story

_No response_

andypalmi:
## Summary
- Adds `automation/add-nodes` action to `ExpertAutomations`
- Mirrors `RED.nodes.importNodes()` internals: sets `_def`, `inputs`, `outputs`, populates `_config` to prevent the orange misconfigured triangle
- Handles dynamic inputs/outputs (e.g. function node)
- Validates nodes via `RED.editor.validateNode()`

Closes #196

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -278,6 +278,13 @@ export class ExpertAutomations extends ExpertActionsInterface {
             }
             return node
         })
+        // importNodes places nodes on the active workspace when addFlow=false,
+        // so switch to the target tab first if nodes target a different one
+        const targetZ = prepared[0]?.z
+        const activeZ = this.RED.workspaces.active()
+        if (targetZ && targetZ !== activeZ) {
+            this.RED.workspaces.show(targetZ)
+        }
         this.RED.view.importNodes(prepared, { generateIds: false, addFlow: false, notify: false, touchImport: true })
         this.RED.nodes.dirty(true)
     }
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -442,17 +442,12 @@ export class ExpertAutomations extends ExpertActionsInterface {
             if (!def) throw new Error(`Unknown node type: ${rawNode.type}`)
             return { ...rawNode }
         })
-        // Validate target tab exists before switching
-        const targetZ = prepared[0]?.z
-        if (targetZ) {
-            const targetWs = this.RED.nodes.workspace(targetZ)
-            if (!targetWs) throw new Error(`Target tab ${targetZ} not found`)
-        }
-        // importNodes places nodes on the active workspace when addFlow=false,
-        // so switch to the target tab first if nodes target a different one
-        const activeZ = this.RED.workspaces.active()
-        if (targetZ && targetZ !== activeZ) {
-            this.showWorkspace(targetZ)
+        // Validate all target tabs exist and are not locked
+        const uniqueZs = [...new Set(prepared.map(n => n.z))]
+        for (const z of uniqueZs) {
+            this.showWorkspace(z) // validates workspace exists (throws if not)
+            const ws = this.RED.nodes.workspace(z)
+            if (ws.locked) throw new Error(`Target tab ${z} is locked`)
         }
         this.RED.view.importNodes(prepared, { generateIds, addFlow: false, notify: false, touchImport: true, applyNodeDefaults: true })
         // Validate import actually succeeded (only when IDs are known)
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -261,22 +261,14 @@ export class ExpertAutomations extends ExpertActionsInterface {
      * @param {Object[]} nodes - array of raw node objects (must include id, type, z)
      */
     addNodes (nodes) {
-        // Validate required fields and types, then fill in default values for omitted properties
+        // Validate required fields and types
         const prepared = nodes.map(rawNode => {
             if (!rawNode.id) throw new Error('Node is missing required property: id')
             if (!rawNode.type) throw new Error('Node is missing required property: type')
             if (!rawNode.z) throw new Error('Node is missing required property: z')
             const def = this.RED.nodes.getType(rawNode.type)
             if (!def) throw new Error(`Unknown node type: ${rawNode.type}`)
-            const node = { ...rawNode }
-            if (def.defaults) {
-                for (const d in def.defaults) {
-                    if (Object.prototype.hasOwnProperty.call(def.defaults, d) && !Object.prototype.hasOwnProperty.call(node, d)) {
-                        node[d] = def.defaults[d].value
-                    }
-                }
-            }
-            return node
+            return { ...rawNode }
         })
         // importNodes places nodes on the active workspace when addFlow=false,
         // so switch to the target tab first if nodes target a different one
@@ -285,7 +277,7 @@ export class ExpertAutomations extends ExpertActionsInterface {
         if (targetZ && targetZ !== activeZ) {
             this.RED.workspaces.show(targetZ)
         }
-        this.RED.view.importNodes(prepared, { generateIds: false, addFlow: false, notify: false, touchImport: true })
+        this.RED.view.importNodes(prepared, { generateIds: false, addFlow: false, notify: false, touchImport: true, applyNodeDefaults: true })
         this.RED.nodes.dirty(true)
     }
 
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -256,58 +256,27 @@ export class ExpertAutomations extends ExpertActionsInterface {
 
     /**
      * Add one or more nodes to the live NR4 canvas.
+     * Delegates to RED.view.importNodes which handles node initialisation,
+     * history (undo/redo) and view updates internally.
      * @param {Object[]} nodes - array of raw node objects (must include id, type, z)
      */
     addNodes (nodes) {
-        const addedNodes = []
-        for (const rawNode of nodes) {
+        // Validate types and fill in default values for omitted properties
+        const prepared = nodes.map(rawNode => {
             const def = this.RED.nodes.getType(rawNode.type)
             if (!def) throw new Error(`Unknown node type: ${rawNode.type}`)
-            const node = {
-                id: rawNode.id, type: rawNode.type,
-                x: parseFloat(rawNode.x || 0), y: parseFloat(rawNode.y || 0),
-                z: rawNode.z, name: rawNode.name || '', changed: true,
-                wires: [], _config: {}, _def: def,
-                inputs: def.inputs || 0, outputs: def.outputs || 0
-            }
-            for (const d in def.defaults) {
-                if (Object.prototype.hasOwnProperty.call(def.defaults, d) && d !== 'inputs' && d !== 'outputs') {
-                    node[d] = Object.prototype.hasOwnProperty.call(rawNode, d) ? rawNode[d] : def.defaults[d].value
-                    node._config[d] = JSON.stringify(node[d])
+            const node = { ...rawNode }
+            if (def.defaults) {
+                for (const d in def.defaults) {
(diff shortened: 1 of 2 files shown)


cstns (changes requested):

andypalmi:

### Changes made (5 commits after initial feat commit)

1. **55ad761 -- Add history to addNodes** (original review fix)
2. **4cbad89 -- Fill node type defaults** -- `RED.view.importNodes` does not fill defaults from `RED.nodes.getType().defaults`, causing "Invalid repeat value" errors on inject nodes. Added preprocessing step.
3. **97db61d -- Delegate to importNodes** -- The original approach (manual `RED.nodes.add()` + `RED.history.push()`) caused undo crashes (`Cannot read properties of undefined (reading 'z')`) because manually-added nodes lack internal NR4 properties that undo expects. Refactored to delegate to `RED.view.importNodes()` which handles node initialization, history, and view updates internally. Added `RED.nodes.dirty(true)` since `importNodes` does not set the dirty flag.
4. **2e7afe5 -- Validate required properties** -- Nodes missing `z` were silently added to the active tab by `importNodes`. Now throws early with a clear message.
5. **68d3ac9 -- Switch to target tab** -- `importNodes` with `addFlow: false` places nodes on the active workspace regardless of `z` value. Now switches to the target tab before importing so nodes land on the correct workspace.

### Manual test results

Run in the NR editor iframe console after pasting the `t()` helper:

```js
    try {
    } catch (e) {
    }
    return result
}
```

#### Happy path

```js
// 1. Add single inject node
const z = RED.workspaces.active()
let r = await t('automation/add-nodes', {
    nodes: [{ id: RED.nodes.id(), type: 'inject', x: 200, y: 200, z }]
})
// PASS: r.success === true, node visible, dirty=true, history peek t='add' with 1 node
// Ctrl+Z removes node, Ctrl+Shift+Z restores it

// 2. Add multiple nodes at once
const id1 = RED.nodes.id(), id2 = RED.nodes.id()
r = await t('automation/add-nodes', {
    nodes: [
        { id: id1, type: 'inject', x: 100, y: 300, z },
        { id: id2, type: 'debug', x: 400, y: 300, z }
    ]
})
// PASS: both visible, history peek nodes.length === 2, single Ctrl+Z removes both

// 3. Add node with custom properties
r = await t('automation/add-nodes', {
    nodes: [{
        id: RED.nodes.id(), type: 'inject', x: 200, y: 400, z,
        payload: 'hello', payloadType: 'str', topic: 'test/topic'
    }]
})
// PASS: edit dialog shows payload='hello', topic='test/topic'

// 4. Add node to a different tab
const tabs = []; RED.nodes.eachWorkspace(ws => tabs.push(ws.id))
const otherTab = tabs.find(tid => tid !== RED.workspaces.active())
r = await t('automation/add-nodes', {
    nodes: [{ id: RED.nodes.id(), type: 'debug', x: 100, y: 100, z: otherTab }]
})
// PASS: editor switches to otherTab, node visible on that tab

// 5. Add function node with custom outputs
r = await t('automation/add-nodes', {
    nodes: [{ id: RED.nodes.id(), type: 'function', x: 300, y: 500, z, outputs: 3 }]
})
// PASS: function node has 3 output ports
```

#### Bad path

```js
// 6. Unknown node type
r = await t('automation/add-nodes', {
    nodes: [{ id: RED.nodes.id(), type: 'nonexistent-xyz', x: 0, y: 0, z }]
})
// PASS: throws "Unknown node type: nonexistent-xyz"

// 7. Missing required z
r = await t('automation/add-nodes', {
    nodes: [{ id: RED.nodes.id(), type: 'inject', x: 0, y: 0 }]
})
// PASS: throws "Node is missing required property: z"

// 8. Empty nodes array
r = await t('automation/add-nodes', { nodes: [] })
// PASS: no error, importNodes creates empty history entry (acceptable)
```

### Unit test coverage

161 tests passing. Added tests for: missing required property validation (id, type, z), tab switching when targeting different workspace, no tab switch when targeting active workspace.

resources/expertAutomations.js
```diff
@@ -229,6 +254,59 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+                node.inputs = parseInt(rawNode.inputs, 10) || def.inputs || 0
+                node._config.inputs = JSON.stringify(rawNode.inputs)
+            }
+            node._ = def._
+            this.RED.nodes.add(node)
+            if (this.RED.editor?.validateNode) {
+                this.RED.editor.validateNode(node)
+            }
+        }
+        this.RED.nodes.dirty(true)
```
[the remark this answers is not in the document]
andypalmi: Fixed this issue in another way.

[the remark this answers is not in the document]
andypalmi: And integrated history functionality

andypalmi:
@cstns  History now works, nodes are validated and any node's property not passed is set to the node's default values for it to avoid any "Invalid properties: prop1, prop2, ..." errors in Node-RED about a property value.

cc: @Steve-Mcl @allthedoll @dimitrieh

andypalmi:
## Update: delegate node defaults to NR4 `importNodes`

Commit 1bf6dbe removes the manual default-filling loop (`def.defaults` iteration) from `addNodes` and instead passes `applyNodeDefaults: true` to `RED.view.importNodes`. This is the same mechanism NR4 uses internally for clipboard paste — it fills missing properties from the node type definition during import.

This simplifies `addNodes` (removed 8 lines of manual default logic) while achieving the same result. Re-tested in the NR editor console — all checks pass, no more "Invalid properties" warnings.

@cstns @Steve-Mcl @allthedoll @dimitrieh

andypalmi:
Added `additionalProperties: true` to the node items schema (e737cf2).

This allows type-specific node properties (name, func, topic, rules, etc.) to pass through, since they are top-level properties on the node object passed directly to `importNodes`.

`x` and `y` remain optional (not required) because config nodes and subflows don't have canvas coordinates.

cstns (approved):

resources/expertAutomations.js
```diff
@@ -98,6 +99,31 @@ export class ExpertAutomations extends ExpertActionsInterface {
                 }
             }
         }
+,
```
Steve-Mcl: clearly linting is not working correctly on this project.

I will raise a PR to fix linting before we go through all these PRs

Steve-Mcl (changes requested):
please fix all linting errors so we can get this merged.

resources/expertAutomations.js
Steve-Mcl: linting was fixed in PR #255 - this has revealed a number of issue that require fixing.

resources/expertAutomations.js
```diff
@@ -229,6 +254,33 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+            const def = this.RED.nodes.getType(rawNode.type)
+            if (!def) throw new Error(`Unknown node type: ${rawNode.type}`)
+            return { ...rawNode }
+        })
+        // importNodes places nodes on the active workspace when addFlow=false,
+        // so switch to the target tab first if nodes target a different one
+        const targetZ = prepared[0]?.z
+        const activeZ = this.RED.workspaces.active()
+        if (targetZ && targetZ !== activeZ) {
+            this.RED.workspaces.show(targetZ)
```
Steve-Mcl: Not sure how important this is right now but if the target tab does not exist, this will not fail and therefore the import will occur on the users current tab.

resources/expertAutomations.js
```diff
@@ -229,6 +254,33 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+        }
+        this.RED.view.importNodes(prepared, { generateIds: false, addFlow: false, notify: false, touchImport: true, applyNodeDefaults: true })
```
Steve-Mcl: Ideally, these options should also be passable from Expert.

Their default value can be as per the PR but should be able to set them via API (for extensibility) 


Reason 1: Action Links could use this to "add" a node
Reason 2: If the ANY ONE of node IDs already exist, the import will silently fail for all nodes. Being able to set `generateIds` will permit the flow import to succeed.

resources/expertAutomations.js
```diff
@@ -229,6 +254,33 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+        this.RED.nodes.dirty(true)
```
Steve-Mcl, line 460: If the import fails (see reasons above) and nothing changes, we are setting the workspace dirty unconditionally (i.e. there is no check that anything actually happened)

Steve-Mcl (commented):
@andypalmi Mostly good however there are some gotchas I want to point out.  If operations for editors are sequential and one fails, we could leave the users workspace in a bit of a mess.

For example, because `generateIds: false`, if a node ID already exists in the flow, the importNodes will silently fail & we dont validate that it actually occurred.

It may be wise to validate the items were placed on the workspace and throw a suitable error for the LLM to iterate on if not?

In the same regard, we dont validate if the workspace ID actually exists or if the switch to that TAB actually occurred - that could mean items are imported to wrong tab (and thus make a mess of users workspace)

In the interest of getting this right, perhaps we should throw a suitable error for this too?

Steve-Mcl:
@andypalmi can you resolve merge conflicts please?

andypalmi:
@Steve-Mcl I worked on it in commit [542dd62](https://github.com/FlowFuse/nr-assistant/pull/235/commits/542dd626107c282dcef1107477b46e593421b9b4) , can you do a last check if the changes look good? Thanks

resources/expertAutomations.js
```diff
@@ -398,6 +424,47 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+            const def = this.RED.nodes.getType(rawNode.type)
+            if (!def) throw new Error(`Unknown node type: ${rawNode.type}`)
+            return { ...rawNode }
+        })
+        // Validate target tab exists before switching
+        const targetZ = prepared[0]?.z
+        if (targetZ) {
+            const targetWs = this.RED.nodes.workspace(targetZ)
+            if (!targetWs) throw new Error(`Target tab ${targetZ} not found`)
+        }
```
Steve-Mcl: This assumes all `z` in `nodes` array are the same.

I can see a future where the nodes array would contain contain multiple flows for multiple tabs at some future point. Even if not, we need to validate all `z` have existing tab/workspace.

This should be covered in unit test too. e.g. try to send a set of nodes where [0].z exists and [1].z does not - should throw before making any partial flow changes.

My suggestion here is generate a unique set of `z`, validate & they all exist and are NOT locked before attempting any imports.

Steve-Mcl (changes requested):

resources/expertAutomations.js
Steve-Mcl: Alternatively, for the short term, at least ensure all elements of `nodes` have same `z`

Steve-Mcl (approved):

PR #236, opened 2026-04-02
feat: add automation/remove-nodes action

### LINKED ISSUE #197: Support for Remove Nodes action
### Support for Remove Nodes action
**Action:** `automation/remove-nodes`

Add support for the `automation/remove-nodes` action in `ExpertAutomations`.

This action removes one or more nodes from the live NR4 canvas by ID.

- Resolve each ID via `RED.nodes.node(id)` and call `RED.nodes.remove(node)`
- Call `RED.nodes.dirty(true)`
- Emit `workspace:change` to refresh `activeNodes` and `activeLinks` so removed nodes and their wires disappear from the view

**Params:** `{ ids: string[] }`

### Epic/Story

_No response_

andypalmi:
## Summary
- Adds `automation/remove-nodes` action to `ExpertAutomations`
- Removes nodes by ID via `RED.nodes.remove()`, skips missing nodes
- Marks flow dirty and emits `workspace:change` to refresh canvas

Closes #197


cstns (changes requested):

cstns (changes requested):
we

cstns (approved):

andypalmi:

- **e0a9142 -- Validate all node IDs exist before removing** -- Previously, non-existent IDs were silently skipped, allowing partial removals. Now validates all IDs upfront and throws if any are missing, listing the invalid IDs. This is a separate commit so it can be reverted independently if the behavior needs to change.

### Manual test results

```js
    try {
    } catch (e) {
    }
    return result
}
```

#### Happy path

```js
// 1. Remove single node (use an existing node ID from your canvas)
let r = await t('automation/remove-nodes', { ids: ['<NODE_ID>'] })
// PASS: r.success === true, node gone from canvas
// PASS: RED.history.peek().t === 'delete', RED.history.peek().nodes.length === 1
// UNDO: Ctrl+Z -- node reappears. Ctrl+Shift+Z -- node removed again.

// 2. Remove multiple nodes
r = await t('automation/remove-nodes', { ids: ['<NODE_ID_1>', '<NODE_ID_2>'] })
// PASS: both nodes gone
// PASS: RED.history.peek().nodes.length === 2
// UNDO: Ctrl+Z -- both reappear in one step

// 3. Remove wired node -- links collected for undo
// (pick a node that has wires connected to it)
r = await t('automation/remove-nodes', { ids: ['<WIRED_NODE_ID>'] })
// PASS: node and its wires removed
// UNDO: Ctrl+Z -- node AND wires both restored
```

#### Bad path

```js
// 4. Non-existent node ID -- throws error
r = await t('automation/remove-nodes', { ids: ['does-not-exist'] })
// PASS: throws "Node(s) not found: does-not-exist"

// 5. Mix of valid and invalid IDs -- throws without removing anything
r = await t('automation/remove-nodes', { ids: ['does-not-exist', '<VALID_NODE_ID>'] })
// PASS: throws "Node(s) not found: does-not-exist"
// PASS: the valid node is NOT removed (pre-check prevents partial operations)

// 6. Empty IDs array -- no-op
r = await t('automation/remove-nodes', { ids: [] })
// PASS: r.success === true, no error, no history entry
```

### Unit test coverage

158 tests passing. Tests cover: remove with history, link collection for undo, throw on non-existent ID, throw on mixed valid/invalid IDs (no partial removal).

andypalmi:
@cstns  If you want you can review it again but I made small changes to make it work using by using IDs, having history work and added tests to perform in browser console to confirm. I tested all cases and it works. 

cc: @Steve-Mcl @allthedoll @dimitrieh

resources/expertAutomations.js
```diff
@@ -461,6 +475,32 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+    /**
+     * Remove one or more nodes from the live NR4 canvas by ID.
+     * @param {string[]} ids - node IDs to remove
+     */
+    removeNodes (ids) {
+        // Validate all IDs exist before removing anything to avoid partial operations
+        const notFound = ids.filter(id => !this.RED.nodes.node(id))
+        if (notFound.length > 0) {
+            throw new Error(`Node(s) not found: ${notFound.join(', ')}`)
+        }
```
Steve-Mcl: Rather than scan all nodes twice, might be a better idea to resolve them once then check for any missing nodes. (line 491 does a 2nd resolve)

resources/expertAutomations.js
```diff
@@ -461,6 +475,32 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+        const allRemovedNodes = []
+        const allRemovedLinks = []
+        for (const id of ids) {
+            const node = this.RED.nodes.node(id)
+            allRemovedNodes.push(node)
+            const removed = this.RED.nodes.remove(id)
```
Steve-Mcl: Should check if the workspace is locked before removing nodes

Steve-Mcl (commented):
1 suggestion & 1 recommendation - we should be checking that the workspace is not locked  - throw appropriate error if the nodes workspace is actually locked.  Since this is an arbitrary array of nodes, there may be many different `z` values (and many same `z` values) so I would suggest grepping a unique set of workspace tab IDs first, verify none of them are locked, then proceed if all OK.

Steve-Mcl (approved):

Steve-Mcl (approved):

Steve-Mcl (approved):

PR #238, opened 2026-04-02
feat: add automation/set-wires action

### LINKED ISSUE #199: Support for Set Wires action
### Description

Action: `automation/set-wires`. Add or remove a single wire between two nodes.

NR4 stores wires as link objects (`{source, sourcePort, target}`), not as `node.wires` arrays. Setting `node.wires` directly has no effect on the canvas.

**Params:**
- `mode` (`'add'` | `'remove'`) — whether to add or remove the wire
- `from` (`string`) — source node ID
- `output` (`number`, optional, default `0`) — source output port index (0-based)
- `to` (`string`) — target node ID

Required: `mode`, `from`, `to`

**Behavior:**
- **add**: resolves both source and target nodes, calls `RED.nodes.addLink({ source, sourcePort, target })`
- **remove**: finds matching link via `RED.nodes.getNodeLinks(from)` filtered by `sourcePort` and `target.id`, calls `RED.nodes.removeLink(link)`
- Sets `sourceNode.dirty = true`, marks flow dirty, emits `workspace:change` to refresh `activeLinks`

### Epic/Story

_No response_

andypalmi:
## Summary
- Adds `automation/set-wires` action to `ExpertAutomations`
- Delta add/remove mode for individual wires
- Uses NR4 link objects (`RED.nodes.addLink`/`removeLink`) — setting `node.wires` directly has no effect

Closes #199


resources/expertAutomations.js
```diff
@@ -229,6 +243,35 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+            this.RED.nodes.addLink({ source: sourceNode, sourcePort: port, target: targetNode })
+        } else {
+            const existingLinks = this.RED.nodes.getNodeLinks(from)
+            const link = existingLinks.find(l =>
+                l.source?.id === from && l.sourcePort === port && l.target?.id === to
+            )
+            if (link) { this.RED.nodes.removeLink(link) }
+        }
+        sourceNode.dirty = true
+        this.RED.nodes.dirty(true)
```
cstns, line 572: test do not react to nodes not being set to dirty

resources/expertAutomations.js
```diff
@@ -229,6 +243,35 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+        if (mode === 'add') {
+            const targetNode = this.RED.nodes.node(to)
+            if (!targetNode) throw new Error(`Target node ${to} not found`)
+            this.RED.nodes.addLink({ source: sourceNode, sourcePort: port, target: targetNode })
+        } else {
+            const existingLinks = this.RED.nodes.getNodeLinks(from)
+            const link = existingLinks.find(l =>
+                l.source?.id === from && l.sourcePort === port && l.target?.id === to
+            )
+            if (link) { this.RED.nodes.removeLink(link) }
```
cstns: tests do not react if tests do not remove a link between nodes

resources/expertAutomations.js
```diff
@@ -229,6 +243,35 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+     * @param {string} params.from - Source node ID
+     * @param {number} [params.output] - Source output port index (0-based, defaults to 0)
+     * @param {string} params.to - Target node ID
+     */
+    setWires ({ mode, from, output, to }) {
+        const sourceNode = this.RED.nodes.node(from)
+        if (!sourceNode) throw new Error(`Source node ${from} not found`)
+        const port = output ?? 0
+        if (mode === 'add') {
+            const targetNode = this.RED.nodes.node(to)
```
cstns: I was able to link two nodes and remove the link between two nodes, nice job!

cstns: I feel like we should look into how to make it clearer to the user that a node was changed and provide undo functionality

cstns (changes requested):
looking good, only concern is with thests not reacting fully, see comments above

resources/expertAutomations.js
cstns: could we use `this.RED.history.push` like in https://github.com/FlowFuse/nr-assistant/pull/237/changes?

andypalmi:

- **ba7cd02 -- Throw when removing a wire that does not exist** -- Previously silently did nothing. Now throws with a clear message including source, port, and target IDs. Separate commit for easy revert if behavior needs to change.

### Manual test results

```js
    try {
    } catch (e) {
    }
    return result
}
```

#### Happy path: add wire

```js
// Use existing node IDs from your canvas. Need at least 2 nodes (e.g. inject + debug).

// 1. Add wire with default output 0
let r = await t('automation/set-wires', { mode: 'add', from: '<SOURCE_ID>', to: '<TARGET_ID>' })
// PASS: r.success === true, wire visible on canvas
// PASS: RED.nodes.node('<SOURCE_ID>').changed === true (blue dot)
// PASS: RED.history.peek().t === 'add', RED.history.peek().links.length === 1
// UNDO: Ctrl+Z -- wire disappears. Ctrl+Shift+Z -- wire reappears.

// 2. Add wire with explicit output port
r = await t('automation/set-wires', { mode: 'add', from: '<SOURCE_ID>', output: 0, to: '<TARGET_ID_2>' })
// PASS: second wire visible

// 3. Add wire on non-zero output (use a function node with multiple outputs)
r = await t('automation/set-wires', { mode: 'add', from: '<FUNCTION_NODE_ID>', output: 1, to: '<TARGET_ID>' })
// PASS: wire from port 1 visible
```

#### Happy path: remove wire

```js
// 4. Remove existing wire (first add one if needed)
r = await t('automation/set-wires', { mode: 'remove', from: '<SOURCE_ID>', output: 0, to: '<TARGET_ID>' })
// PASS: wire gone
// PASS: RED.history.peek().t === 'delete', RED.history.peek().links.length === 1
// UNDO: Ctrl+Z -- wire restored
```

#### Bad path

```js
// 5. Remove wire that does not exist -- throws
r = await t('automation/set-wires', { mode: 'remove', from: '<SOURCE_ID>', output: 0, to: 'does-not-exist' })
// PASS: throws "Wire not found from <SOURCE_ID> port 0 to does-not-exist"

// 6. Non-existent source node -- throws
r = await t('automation/set-wires', { mode: 'add', from: 'does-not-exist', to: '<TARGET_ID>' })
// PASS: throws "Source node does-not-exist not found"

// 7. Non-existent target node -- throws
r = await t('automation/set-wires', { mode: 'add', from: '<SOURCE_ID>', to: 'does-not-exist' })
// PASS: throws "Target node does-not-exist not found"
```

### Unit test coverage

159 tests passing. Tests cover: add wire with history, remove wire with history, non-zero output port, throw on missing source, throw on removing non-existent wire.

cstns (changes requested):
undo works perfectly when adding and removing wires and nodes are marked as dirty accordingly. One small niggle, when undo-ing the node is still left marked as dirty, we should fix that

andypalmi:
@cstns NR's expected behaviour does not show a node as changed after changing its wires. I fixed the fact that it was marked as changed so that it's not anymore and the behaviour you described should not happen again.

andypalmi:
**Question: should we validate the output port index?**

Currently `set-wires` allows wiring on an output port that doesn't exist on the source node. For example, calling `set-wires` with `output: 1` on a node that has `outputs: 1` (i.e. only port 0) silently creates the wire.

We could validate `port < sourceNode.outputs` and throw if the port is out of range. Should we add this check, or is there a reason to allow it (e.g. dynamic outputs that get added later)?

Steve-Mcl (approved):

Steve-Mcl (commented):
Code does not check the following states:
1. Flow/TAB is locked.
1. Calling add wire where a wire already exists & then sets the flow `dirty` unconditionally.
1. Calling remove wire where a wire does not already exists & then sets the flow `dirty` unconditionally.

I would propose 1 of 2 directions:
Either throw an error if the expected links already exists/dont exist OR we only mark dirty/update/redraw IF the state actually changes. 
For locked workspaces, we should probably check if the ws is locked first and throw an appropriate error for the AI to interpret.

Steve-Mcl (changes requested):
On re-review I see the following issues:
1. Does not check the node can accept inputs
2. Does not check the node has the specified output port
3. Does not check the workspace tab is locked
4. Does not check the source and target are on the same tab
5. Adds a wire connection to a config node
6. Allows invalid node loop

Image

Image

andypalmi:
## Manual Testing Results — Validation Checks

Tested the latest changes against the issues raised in review. All validations are working correctly.

### Happy path
- **Add wire** between two nodes: `success: true`
- **Remove wire** between two nodes: `success: true`

### Node existence checks
- Add wire from non-existent source node: `"Source node not found"`
- Add wire to non-existent target node: `"Target node not found"`
- Remove wire to non-existent target node: `"Target node not found"`

### Output port validation
- Add wire specifying an output port that does not exist on a single-output node: `"Source node does not have output port"`
- Add wire specifying an out-of-range output port: `"Source node does not have output port"`
- Add wire using a node with 0 outputs as source: `"Source node does not have output port 0"`

### Workspace locked check
- Add wire on a locked workspace: `"Cannot modify wires — workspace is locked"`

### Same tab check
- Add wire between nodes on different tabs: `"Source and target nodes must be on the same tab"`

### Self-wiring check
- Wire a node to itself: `"Cannot wire a node to itself"`

### Duplicate / missing wire checks
- Add a wire that already exists: `"Wire already exists"`
- Adding a wire and then adding the same wire again correctly rejects the duplicate

@Steve-Mcl

Steve-Mcl (approved):

PR #239, opened 2026-04-02
feat: add automation/add-tab action

### LINKED ISSUE #200: Support for Add Tab action
### Support for Add Tab action
**Action:** `automation/add-tab`

Add support for the `automation/add-tab` action in `ExpertAutomations`.

This action creates a new flow tab in the NR4 editor with an explicit configuration. If no `id` is provided, one is auto-generated via `RED.nodes.id()`.

Two internal calls are required — `RED.workspaces.add(ws)` alone is insufficient:
- It only calls `workspace_tabs.addTab()` (tab bar UI) but skips `RED.nodes.addWorkspace()`
- Without the nodes model registration, `RED.nodes.workspace(id)` returns `null` on the next `workspace:change` event
- This causes the view module to set `activeFlowLocked = true`, which silently breaks wire rendering on the new tab

The implementation must call both in order:
1. `RED.nodes.addWorkspace(ws)` — registers in the nodes model
2. `RED.workspaces.add(ws)` — adds to the tab bar UI

**Params:** `{ id?: string, label: string, disabled?: boolean, info?: string, env?: any[] }`

**Note on `automation/add-flow-tab`:** A separate `automation/add-flow-tab` action exists (added in #171) which goes through NR's `core:add-flow` action and uses `commandAndWait` to wait for the `flows:add` event. It does not accept an `id` — NR auto-generates it. That action is marked `@deprecated` in favour of `automation/add-tab`, which now also auto-generates an id when one is not provided, covering both use cases.

### Epic/Story

_No response_

andypalmi:
## Summary
- Adds `automation/add-tab` action to `ExpertAutomations`
- Calls both `RED.nodes.addWorkspace()` and `RED.workspaces.add()` to avoid the `activeFlowLocked` wire rendering bug
- Supports explicit tab ID or auto-generation
- Deprecates `automation/add-flow-tab`

Closes #200

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -98,8 +98,7 @@ export class ExpertAutomations extends ExpertActionsInterface {
                     }
                 }
             }
-        }
-,
+        },
         [ADD_TAB]: {
             params: {
                 type: 'object',
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -355,6 +355,9 @@ export class ExpertAutomations extends ExpertActionsInterface {
      */
     addTab (tab) {
         if (tab.label == null) throw new Error('Tab label is required')
+        if (tab.id && (this.RED.nodes.node(tab.id) || this.RED.nodes.workspace(tab.id) || this.RED.nodes.subflow(tab.id))) {
+            throw new Error(`ID ${tab.id} already exists — provide a unique ID or omit to auto-generate`)
+        }
         const ws = {
             type: 'tab',
             id: tab.id || this.RED.nodes.id(),
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -6,9 +6,10 @@ const GET_NODES = 'automation/get-nodes'
 const EDIT_NODE = 'automation/open-node-edit'
 const SEARCH = 'automation/search'
 const ADD_FLOW_TAB = 'automation/add-flow-tab'
+const ADD_TAB = 'automation/add-tab'
 
 /**
- * @typedef {SELECT_NODES|GET_NODES|EDIT_NODE|SEARCH|ADD_FLOW_TAB} ExpertAutomationsActionsEnum
+ * @typedef {SELECT_NODES|GET_NODES|EDIT_NODE|SEARCH|ADD_FLOW_TAB|ADD_TAB} ExpertAutomationsActionsEnum
  */
 
 export class ExpertAutomations extends ExpertActionsInterface {
@@ -96,6 +97,20 @@ export class ExpertAutomations extends ExpertActionsInterface {
                         description: 'Optional title for the new flow tab'
                     }
                 }
+            }
+        }
+,
+        [ADD_TAB]: {
+            params: {
+                type: 'object',
+                properties: {
+                    id: { type: 'string', description: 'Tab ID — auto-generated if omitted' },
+                    label: { type: 'string', description: 'Tab label' },
+                    disabled: { type: 'boolean', description: 'Create as disabled' },
+                    info: { type: 'string', description: 'Tab notes' },
+                    env: { type: 'array', description: 'Environment variables' }
+                },
+                required: ['label']
             }
         }
     })
@@ -229,6 +244,24 @@ export class ExpertAutomations extends ExpertActionsInterface {
         return newTab
     }
 
+    /**
+     * Add a new flow tab with an explicit ID and configuration.
+     * @param {Object} tab - tab definition with id, label, disabled, info, env
+     */
+    addTab (tab) {
+        const ws = {
+            type: 'tab',
+            id: tab.id || this.RED.nodes.id(),
+            label: tab.label,
+            disabled: tab.disabled || false,
+            info: tab.info || '',
+            env: tab.env || []
+        }
+        this.RED.nodes.addWorkspace(ws)
+        this.RED.workspaces.add(ws)
+        this.RED.workspaces.show(ws.id)
+    }
+
     get supportedActions () {
         return this.actions
     }
@@ -300,6 +333,10 @@ export class ExpertAutomations extends ExpertActionsInterface {
         }
             break
 
+        case ADD_TAB:
+            this.addTab(params)
+            result.success = true
+            break
         default:
             result.handled = false
             result.success = false
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -260,6 +260,7 @@ export class ExpertAutomations extends ExpertActionsInterface {
      * @param {Object} tab - tab definition with id, label, disabled, info, env
      */
     addTab (tab) {
+        if (tab.label == null) throw new Error('Tab label is required')
         const ws = {
             type: 'tab',
             id: tab.id || this.RED.nodes.id(),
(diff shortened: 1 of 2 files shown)


resources/expertAutomations.js
```diff
@@ -229,6 +244,24 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+            type: 'tab',
+            id: tab.id || this.RED.nodes.id(),
+            label: tab.label,
+            disabled: tab.disabled || false,
+            info: tab.info || '',
+            env: tab.env || []
+        }
+        this.RED.nodes.addWorkspace(ws)
+        this.RED.workspaces.add(ws)
+        this.RED.workspaces.show(ws.id)
```
cstns, line 373: the workspace is not marked as dirty and can't deploy flows after a tab as been created in order to save the new tab

resources/expertAutomations.js
```diff
@@ -98,6 +99,20 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+,
+        [ADD_TAB]: {
+            params: {
+                type: 'object',
+                properties: {
+                    id: { type: 'string', description: 'Tab ID — auto-generated if omitted' },
+                    label: { type: 'string', description: 'Tab label' },
+                    disabled: { type: 'boolean', description: 'Create as disabled' },
+                    info: { type: 'string', description: 'Tab notes' },
+                    env: { type: 'array', description: 'Environment variables' }
```
cstns: maybe we should type-hint this better to make it clear that env is an array of objects with keys name,value type

resources/expertAutomations.js
```diff
@@ -229,6 +244,24 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+        const ws = {
+            type: 'tab',
+            id: tab.id || this.RED.nodes.id(),
+            label: tab.label,
+            disabled: tab.disabled || false,
+            info: tab.info || '',
+            env: tab.env || []
+        }
+        this.RED.nodes.addWorkspace(ws)
+        this.RED.workspaces.add(ws)
```
cstns, line 370: tests do not react to not adding the workspace using this method

resources/expertAutomations.js
cstns: also, tests do not react when the current workspace is not selected after the addition of a new one

resources/expertAutomations.js
```diff
@@ -229,6 +244,24 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+     * @param {Object} tab - tab definition with id, label, disabled, info, env
+     */
+    addTab (tab) {
+        const ws = {
+            type: 'tab',
+            id: tab.id || this.RED.nodes.id(),
+            label: tab.label,
+            disabled: tab.disabled || false,
+            info: tab.info || '',
+            env: tab.env || []
```
cstns, line 367: tests do not react when the disabled, info and env keys are not given, we should add test scenarios to cover them

cstns (changes requested):

cstns (approved):

andypalmi:

- **43284f7 -- Validate that label is provided** -- Missing label caused NR's `addWorkspace` to crash with `Cannot read properties of undefined (reading 'indexOf')`. Now throws early with "Tab label is required". Separate commit for easy revert.

### Manual test results

```js
    try {
    } catch (e) {
    }
    return result
}
```

#### Happy path

```js
// 1. Add tab with label only
let r = await t('automation/add-tab', { label: 'Test Tab 1' })
// PASS: r.success === true, tab visible in tab bar, editor switched to it
// PASS: RED.history.peek().t === 'add', RED.history.peek().workspaces.length === 1
// UNDO: Ctrl+Z -- tab disappears from tab bar

// 2. Add tab with all options
r = await t('automation/add-tab', {
    label: 'Full Options Tab',
    disabled: true,
    info: 'These are tab notes',
    env: [{ name: 'MY_VAR', value: 'hello', type: 'str' }]
})
// PASS: tab appears disabled (striped/greyed in tab bar)
// PASS: double-click tab -- edit shows info='These are tab notes', env has MY_VAR

// 3. Add tab with disabled=false explicitly
r = await t('automation/add-tab', { label: 'Enabled Tab', disabled: false })
// PASS: tab is active (not greyed)

// 4. Add tab with multiple env vars
r = await t('automation/add-tab', {
    label: 'Env Tab',
    env: [
        { name: 'API_KEY', value: 'secret123', type: 'str' },
        { name: 'DEBUG', value: 'true', type: 'bool' }
    ]
})
// PASS: double-click tab -- env has 2 variables
```

#### Bad path

```js
// 5. Missing required label -- throws
r = await t('automation/add-tab', {})
// PASS: throws "Tab label is required"

// 6. Empty label string -- NR allows it
r = await t('automation/add-tab', { label: '' })
// PASS: tab created with empty label (NR permits this)
```

### Unit test coverage

157 tests passing. Tests cover: tab creation with history+dirty, default values for omitted fields, throw on missing label.

resources/expertAutomations.js
```diff
@@ -341,6 +366,27 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
 
+    /**
+     * Add a new flow tab with an explicit ID and configuration.
+     * @param {Object} tab - tab definition with id, label, disabled, info, env
+     */
+    addTab (tab) {
+        if (tab.label == null) throw new Error('Tab label is required')
+        const ws = {
+            type: 'tab',
+            id: tab.id || this.RED.nodes.id(),
```
Steve-Mcl, line 363: If `tab.id` is a duplicate, artifacts can occur:

Image

I recommend that the `tab.id` - if provided, MUST not be an existing id (node/subflow/tab/anything!) - it should throw an error!

Steve-Mcl (changes requested):
duplicate tab id needs to be handled.

resources/expertAutomations.js
```diff
@@ -324,6 +349,30 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+        if (tab.id && (this.RED.nodes.node(tab.id) || this.RED.nodes.workspace(tab.id) || this.RED.nodes.subflow(tab.id))) {
+            throw new Error(`ID ${tab.id} already exists — provide a unique ID or omit to auto-generate`)
+        }
+        const ws = {
+            type: 'tab',
+            id: tab.id || this.RED.nodes.id(),
+            label: tab.label,
+            disabled: tab.disabled || false,
+            info: tab.info || '',
+            env: tab.env || []
```
Steve-Mcl, line 367: I think it is time we improved (or replaced) our super basic schema validation.
In our basic validateSchema function, the `env` array will accept absolutely anything.

e.g. 
```
{
                    type: 'invoke-action',
                    action: 'automation/add-tab',
                    params: {
                        label: 'p',
                        disabled: true,
                        info: '',
                        env: [{ id: 'qwe1', value: '123', type: 'integer' }]
                    }
                }
```

looks valid - however there are 2 issues.
1. it should be `name` not `id`
2. type cannot be `integer`. Supported types for a tab (or group) `env` are str, num, bool, json, env, cred, jsonata

I will not block this PR for that but will raise a follow up issue that we need to implement before allowing AI to run wild with this.

Steve-Mcl (approved):
Approved with follow up task https://github.com/FlowFuse/nr-assistant/issues/258

resources/expertAutomations.js
Steve-Mcl: follow up task https://github.com/FlowFuse/nr-assistant/issues/258

PR #241, opened 2026-04-02
feat: add automation/get-workspace-nodes action

### LINKED ISSUE #202: Support for Get Workspace Nodes action (automation/get-workspace-nodes)
### Support for Get Flow action
**Action:** `automation/get-flow`

Add support for the `automation/get-flow` action in `ExpertAutomations`.

This action reads the live canvas state — including undeployed edits that only exist in the editor's in-memory model — and returns it as a flows array. It is the only mechanism that can read undeployed state; the NR runtime API only knows about deployed flows.

**Planned transport:** Will be called on canvas change events (e.g. `nodes:change`, `workspace:change`) and the result pushed over WebSocket to the FlowFuse Expert. Not yet implemented.

**Reading logic:**
- `RED.nodes.eachWorkspace()` — collect tabs
- `RED.nodes.eachNode()` — collect nodes, reconstruct `wires` from link objects via `RED.nodes.getNodeLinks()` (reading `node.wires` directly returns stale/empty data in NR4)
- Copy user-edited values from `node._config` for each property in `node._def.defaults`
- Include config nodes (no `z` field — not scoped to a tab)

**Params:** none
**Returns:** `{ flows: FlowJSON[] }`

### Epic/Story

_No response_

andypalmi:
## Summary
- Adds `automation/get-flow` action to `ExpertAutomations`
- Reads live canvas state including undeployed edits
- Reconstructs wires from NR4 link objects (reading `node.wires` returns stale data)
- Copies user-edited values from `node._config`

Closes #202

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -6,9 +6,10 @@ const GET_NODES = 'automation/get-nodes'
 const EDIT_NODE = 'automation/open-node-edit'
 const SEARCH = 'automation/search'
 const ADD_FLOW_TAB = 'automation/add-flow-tab'
+const GET_FLOW = 'automation/get-flow'
 
 /**
- * @typedef {SELECT_NODES|GET_NODES|EDIT_NODE|SEARCH|ADD_FLOW_TAB} ExpertAutomationsActionsEnum
+ * @typedef {SELECT_NODES|GET_NODES|EDIT_NODE|SEARCH|ADD_FLOW_TAB|GET_FLOW} ExpertAutomationsActionsEnum
  */
 
 export class ExpertAutomations extends ExpertActionsInterface {
@@ -97,6 +98,10 @@ export class ExpertAutomations extends ExpertActionsInterface {
                     }
                 }
             }
+        }
+,
+        [GET_FLOW]: {
+            params: null
         }
     })
 
@@ -229,6 +234,42 @@ export class ExpertAutomations extends ExpertActionsInterface {
         return newTab
     }
 
+    /**
+     * Read the live canvas state (including undeployed edits) and return it.
+     * @returns {Object[]} full flows array (tabs + nodes + config nodes)
+     */
+    getFlow () {
+        const flows = []
+        this.RED.nodes.eachWorkspace(ws => {
+            flows.push({ id: ws.id, type: 'tab', label: ws.label, disabled: ws.disabled || false })
+        })
+        this.RED.nodes.eachNode(node => {
+            const plain = { id: node.id, type: node.type, z: node.z, name: node.name }
+            if (node.x !== undefined) plain.x = node.x
+            if (node.y !== undefined) plain.y = node.y
+            if (node.outputs > 0) {
+                const wires = Array.from({ length: node.outputs }, () => [])
+                this.RED.nodes.getNodeLinks(node.id).forEach(link => {
+                    if (link.source?.id === node.id && wires[link.sourcePort]) {
+                        wires[link.sourcePort].push(link.target.id)
+                    }
+                })
+                plain.wires = wires
+            } else {
+                plain.wires = []
+            }
+            if (node._config) {
+                for (const k of Object.keys(node._config)) {
+                    if (k !== 'x' && k !== 'y' && plain[k] === undefined) {
+                        try { plain[k] = JSON.parse(node._config[k]) } catch { plain[k] = node._config[k] }
+                    }
+                }
+            }
+            flows.push(plain)
+        })
+        return flows
+    }
+
     get supportedActions () {
         return this.actions
     }
@@ -300,6 +341,10 @@ export class ExpertAutomations extends ExpertActionsInterface {
         }
             break
 
+        case GET_FLOW:
+            result.flows = this.getFlow()
+            result.success = true
+            break
         default:
             result.handled = false
             result.success = false
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -6,7 +6,7 @@ const GET_NODES = 'automation/get-nodes'
 const EDIT_NODE = 'automation/open-node-edit'
 const SEARCH = 'automation/search'
 const ADD_FLOW_TAB = 'automation/add-flow-tab'
-const GET_FLOW = 'automation/get-flow'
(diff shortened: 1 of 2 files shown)


resources/expertAutomations.js
```diff
@@ -6,9 +6,10 @@ const GET_NODES = 'automation/get-nodes'
 const EDIT_NODE = 'automation/open-node-edit'
 const SEARCH = 'automation/search'
 const ADD_FLOW_TAB = 'automation/add-flow-tab'
+const GET_FLOW = 'automation/get-flow'
```
cstns: this should be renamed to something more explicit to resemble what it's actually doing, getting all of the nodes from the active flow tab

resources/expertAutomations.js
```diff
@@ -229,6 +234,42 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+        const flows = []
+        this.RED.nodes.eachWorkspace(ws => {
+            flows.push({ id: ws.id, type: 'tab', label: ws.label, disabled: ws.disabled || false })
+        })
+        this.RED.nodes.eachNode(node => {
+            const plain = { id: node.id, type: node.type, z: node.z, name: node.name }
+            if (node.x !== undefined) plain.x = node.x
+            if (node.y !== undefined) plain.y = node.y
+            if (node.outputs > 0) {
+                const wires = Array.from({ length: node.outputs }, () => [])
```
cstns: method not fully covered by coverage

cstns (changes requested):
tested, works as expected but see comments above

cstns (approved):

andypalmi:
## Manual Testing Results (get-workspace-nodes)

**Note:** Action renamed from `automation/get-flow` to `automation/get-workspace-nodes` as requested in review. PR title and issue #202 title updated accordingly.

All tests run in NR editor iframe console on a workspace with 3 tabs and 17 total nodes (including wired function nodes).

### Test Code

```js
const dirtyBefore = RED.nodes.dirty()
const histBefore = RED.history.peek()
let r = await t('automation/get-workspace-nodes', {})
check('1a: success', r.success === true)
check('1b: flows is array', Array.isArray(r.flows))
check('1c: has tab entries', r.flows?.some(f => f.type === 'tab'))
check('1d: dirty unchanged', RED.nodes.dirty() === dirtyBefore)
check('1e: no history entry', RED.history.peek() === histBefore)
console.log('Total items:', r.flows?.length)

const flowTabs = r.flows?.filter(f => f.type === 'tab')
check('2: at least one tab', (flowTabs?.length ?? 0) > 0)

const wiredNodes = r.flows?.filter(f => f.wires?.some(w => w.length > 0))
console.log('Nodes with wires:', wiredNodes?.length)

const funcNodes = r.flows?.filter(f => f.type === 'function')
if (funcNodes?.length > 0) {
    check('4: function node has outputs field', 'outputs' in funcNodes[0])
}

r = await t('automation/get-flow', {})
check('5: old action name fails', r.success === false)
```

### Results

| Test | Result |
|------|--------|
| 1a: success | PASS |
| 1b: flows is array | PASS |
| 1c: has tab entries | PASS |
| 1d: dirty unchanged | PASS |
| 1e: no history entry | PASS |
| 2: at least one tab (3 tabs) | PASS |
| 4: function node has outputs field | PASS |
| 5: old action name fails | PASS |

Total items: 17, tabs: 3, nodes with wires: 7.

Returned node properties include `_config` values (outputs, func, method, url, etc.) and wires are correctly populated from live links.

/cc @cstns @Steve-Mcl @allthedoll @dimitrieh

andypalmi:
## Test Coverage Update

Added 6 unit tests to cover previously uncovered paths in the `getFlow` method:

1. **Node without x/y coordinates** -- nodes that lack x/y (e.g. config nodes) should have those fields omitted from the output rather than showing up as `undefined`

2. **Multiple output ports with wires on different ports** -- a node with `outputs: 2` has two output ports, each with its own connections. The method builds `wires[0]`, `wires[1]`, etc. from link objects. This test verifies links are sorted into the correct port sub-array, e.g. `[['targetA'], ['targetB']]`

3. **`_config` excludes x and y keys** -- `_config` may contain x/y keys, but these should not override the node's actual canvas coordinates already set from the node object itself

4. **`_config` does not overwrite existing plain properties** -- if `_config` contains a key like `name` that is already set on the plain node object, the existing value takes priority

5. **JSON parse failure in `_config` falls back to raw string** -- `_config` values are stored as strings and parsed via `JSON.parse()` (e.g. `'"hello"'` becomes `"hello"`). If the value is not valid JSON (e.g. a raw string like `not-valid-json{`), the catch block falls back to using the raw string as-is instead of throwing

6. **Node without `_config` property** -- some nodes may not have `_config` at all. The method guards with `if (node._config)` -- this test verifies it does not throw and still produces correct output

All 165 tests pass.

/cc @cstns @Steve-Mcl @allthedoll @dimitrieh

resources/expertAutomations.js
```diff
@@ -276,6 +280,42 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+            if (node._config) {
+                for (const k of Object.keys(node._config)) {
+                    if (k !== 'x' && k !== 'y' && plain[k] === undefined) {
+                        try { plain[k] = JSON.parse(node._config[k]) } catch { plain[k] = node._config[k] }
+                    }
+                }
+            }
+            flows.push(plain)
+        })
+        return flows
```
Steve-Mcl: Why are we doing this manually?

A call to `RED.nodes.createCompleteNodeSet({ credentials: false, includeModuleConfig: true });` is what Node-RED export does to get full flows.

We could simply replace all of this with 
```js
return this.RED.nodes.createCompleteNodeSet({ credentials: false, includeModuleConfig: params.includeModuleConfig })
```

Steve-Mcl (commented):
Not sure why we are re-inventing what Node-RED does already?

I may be missing something tho.

Steve-Mcl:
Also, please fix conflicts. Ta.

andypalmi:
See commit [cfad1f1](https://github.com/FlowFuse/nr-assistant/pull/241/commits/cfad1f198a096136a849b3107ca3082a9485385f) @Steve-Mcl

Steve-Mcl (approved):
👍

PR #242, opened 2026-04-02
feat: add automation/import-flow action

### LINKED ISSUE #172: Support for Import Flow action
### Description

As part of seamless integration with Node-RED, we should implement the "Import Flow" action that allows the FlowFuse Expert to provide actionable links for importing flows into the Node-RED editor. This will enable users to quickly bring in new flows based on the expert's recommendations, either by opening the import dialog or by pre-filling it with a provided flow.

Capabilities:
- Open the import dialog in the Node-RED editor.
- Pre-fill the import dialog with a provided flow.

There is some overlap with existing import-flows capability in the nr-assistant however that is specifically for importing a flow the expert got from RAG (without the import dialog)

This issue is about when the Expert responds to questions like "I have this flow, how do i import it"

### Epic/Story

_No response_

andypalmi:
## Summary
- Adds `automation/import-flow` action to `ExpertAutomations`
- Migrates from legacy `custom:import-flow` in expertComms.js (custom: variant kept for backwards compat)
- Supports flow JSON string or array, addFlow flag, and generateIds option
- Adds `touchImport: true` to `importNodes()` call

Closes #172

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -326,7 +326,7 @@ export class ExpertAutomations extends ExpertActionsInterface {
      * @param {boolean} importOptions.addFlow whether to add the nodes to a new flow or to the current flow
      * @param {boolean} [importOptions.notify=true] whether to show notifications for import success/failure (default true)
      */
-    importFlow (nodesStr, { addFlow = false, generateIds = true, notify = true } = { addFlow: false, generateIds: true, notify: true }) {
+    importFlow (nodesStr, { addFlow = false, generateIds = true } = { addFlow: false, generateIds: true }) {
         let newNodes = nodesStr
         if (typeof nodesStr === 'string') {
             try {
@@ -340,10 +340,16 @@ export class ExpertAutomations extends ExpertActionsInterface {
                 e.code = 'NODE_RED'
                 throw e
             }
-        } else if (!Array.isArray(nodesStr)) {
+        } else if (Array.isArray(nodesStr)) {
+            this.redOps.validateFlow(nodesStr)
+        } else {
             throw new Error('importFlow expects a JSON string or an array of node objects')
         }
-        this.RED.view.importNodes(newNodes, { generateIds, addFlow, notify, touchImport: true, applyNodeDefaults: true })
+        // If importing onto the current tab (not creating a new one), check it's not locked
+        if (!addFlow && this.RED.workspaces.isLocked()) {
+            throw new Error('Cannot import into a locked workspace')
+        }
+        this.RED.view.importNodes(newNodes, { generateIds, addFlow, touchImport: true, applyNodeDefaults: true })
         this.RED.nodes.dirty(true)
     }
 
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -203,7 +203,7 @@ export class ExpertAutomations extends ExpertActionsInterface {
     /// Function extracted from Node-RED source `editor-client/src/js/ui/clipboard.js`
     /**
      * Performs the import of nodes, handling any conflicts that may arise
-     * @param {string} nodesStr the nodes to import as a string
+     * @param {string|Object[]} nodesStr the nodes to import — either a JSON string or an array of node objects
      * @param {object} importOptions
      * @param {boolean} importOptions.addFlow whether to add the nodes to a new flow or to the current flow
      * @param {boolean} [importOptions.notify=true] whether to show notifications for import success/failure (default true)
@@ -222,6 +222,8 @@ export class ExpertAutomations extends ExpertActionsInterface {
                 e.code = 'NODE_RED'
                 throw e
             }
+        } else if (!Array.isArray(nodesStr)) {
+            throw new Error('importFlow expects a JSON string or an array of node objects')
         }
         this.RED.view.importNodes(newNodes, { generateIds, addFlow, notify, touchImport: true, applyNodeDefaults: true })
         this.RED.nodes.dirty(true)
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -98,8 +98,7 @@ export class ExpertAutomations extends ExpertActionsInterface {
                     }
                 }
             }
-        }
-,
+        },
         [IMPORT_FLOW]: {
             params: {
                 type: 'object',
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -224,6 +224,7 @@ export class ExpertAutomations extends ExpertActionsInterface {
             }
         }
         this.RED.view.importNodes(newNodes, { generateIds, addFlow, notify, touchImport: true })
+        this.RED.nodes.dirty(true)
     }
 
     async addFlowTab (title) {
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -6,9 +6,10 @@ const GET_NODES = 'automation/get-nodes'
 const EDIT_NODE = 'automation/open-node-edit'
 const SEARCH = 'automation/search'
 const ADD_FLOW_TAB = 'automation/add-flow-tab'
+const IMPORT_FLOW = 'automation/import-flow'
 
 /**
- * @typedef {SELECT_NODES|GET_NODES|EDIT_NODE|SEARCH|ADD_FLOW_TAB} ExpertAutomationsActionsEnum
+ * @typedef {SELECT_NODES|GET_NODES|EDIT_NODE|SEARCH|ADD_FLOW_TAB|IMPORT_FLOW} ExpertAutomationsActionsEnum
  */
 
 export class ExpertAutomations extends ExpertActionsInterface {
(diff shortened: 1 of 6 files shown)


resources/expertAutomations.js
```diff
@@ -98,6 +99,28 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+,
+        [IMPORT_FLOW]: {
+            params: {
+                type: 'object',
+                properties: {
+                    flow: {
+                        type: ['string', 'array'],
+                        description: 'Flow JSON string or array to import onto the canvas'
+                    },
+                    addFlow: {
```
cstns: should be renamed to addFlowTab?

cstns: if we're creating a new flow tab, shouldn't we also be able to parametrize it similarly to the add-tab action pr?

cstns: also, just now finding the discrepancies between pr's on naming flow tabs here it's addFlow and on the previous p it's add-tab we should normalize them

resources/expertAutomations.js
```diff
@@ -98,6 +99,28 @@ export class ExpertAutomations extends ExpertActionsInterface {
                 }
             }
         }
+,
+        [IMPORT_FLOW]: {
+            params: {
+                type: 'object',
+                properties: {
+                    flow: {
+                        type: ['string', 'array'],
```
cstns, line 226: the type allows for both arrays and strings but the importFlow method only accounts for strings

cstns (changes requested):

test/unit/resources/expertAutomations.test.js
```diff
@@ -324,7 +324,7 @@ describeMain('expertAutomations', () => {
             })
         })
             describe('importFlow action', () => {
-                it('should import flow JSON', async () => {
+                it('should import flow JSON string', async () => {
```
cstns: the test was updated but the schema still accepts an array of nodes. we should allow both an array of nodes and a string imo and alter the importFlow method to accomodate

cstns (changes requested):

andypalmi:
## Manual Testing Results — `automation/import-flow`

### Fixes applied

1. **Dirty flag** — `RED.nodes.dirty(true)` now called after `importNodes` (commit cf34475)
2. **Node defaults** — `applyNodeDefaults: true` now passed to `importNodes`, fixing "Invalid repeat value" validation warnings on inject nodes (commit 07cb040)

### Observations (NR4 internal behavior, not bugs in our code)

- **`generateIds: true`** always regenerates IDs (confirmed via NR4 source). `RED.nodes.node(originalId)` returns `undefined` (not `null`) for non-existent nodes — tests must use `== null` (loose equality), not `=== null`.
- **`addFlowTab: true`** does not switch to the new tab. NR4 only auto-switches when it creates a *fallback* workspace for orphaned nodes. When a tab is included in the import payload, nodes map to it directly — no fallback needed, no switch.
- **`generateIds: false`** preserves original IDs as expected. Throws `import_conflict` if a node with the same ID already exists.

### Test code

```js
    try {
    } catch (e) {
    }
    return result
}
const check = (name, ok) => console.log(`${ok ? 'PASS' : 'FAIL'}: ${name}`, ok)

// 1. Import JSON string into current tab (generateIds: true)
RED.nodes.dirty(false)
const beforeCount = RED.nodes.filterNodes({type: 'inject'}).length
let r = await t('automation/import-flow', {
    flow: JSON.stringify([
        { id: 'imp1', type: 'inject', x: 100, y: 100, wires: [['imp2']] },
        { id: 'imp2', type: 'debug', x: 350, y: 100, wires: [] }
    ]),
    generateIds: true
})
const afterCount = RED.nodes.filterNodes({type: 'inject'}).length
check('1a: success', r.success === true)
check('1b: dirty', RED.nodes.dirty() === true)
check('1c: original IDs NOT kept (imp1)', RED.nodes.node('imp1') == null)
check('1c: original IDs NOT kept (imp2)', RED.nodes.node('imp2') == null)
check('1d: new nodes created (IDs regenerated)', afterCount > beforeCount)
console.log('>> MANUAL: Ctrl+Z -> both nodes removed')

// 2. Import as array (not string)
r = await t('automation/import-flow', {
    flow: [
        { id: 'arr1', type: 'inject', x: 100, y: 250, wires: [] }
    ],
    generateIds: true
})
check('2a: success', r.success === true)
check('2b: original ID NOT kept', RED.nodes.node('arr1') == null)

// 3. Import into new tab (addFlowTab=true)
r = await t('automation/import-flow', {
    flow: JSON.stringify([
        { id: 'tab1', type: 'tab', label: 'Imported Tab' },
        { id: 'n1', type: 'inject', x: 100, y: 100, z: 'tab1', wires: [] }
    ]),
    addFlowTab: true,
    generateIds: true
})
check('3a: success', r.success === true)
console.log('>> MANUAL: new "Imported Tab" created (no auto-switch — expected NR4 behavior)')
console.log('>> MANUAL: Ctrl+Z -> tab + node removed')

// 4. Import with generateIds=false -> keeps original IDs
r = await t('automation/import-flow', {
    flow: JSON.stringify([
        { id: 'keep-this-id-999', type: 'inject', x: 200, y: 300, wires: [] }
    ]),
    generateIds: false
})
check('4a: success', r.success === true)
check('4b: ID preserved', RED.nodes.node('keep-this-id-999') != null)

// 5. Import with addFlowTab=false (default) -> current tab
const tabBefore = RED.workspaces.active()
r = await t('automation/import-flow', {
    flow: JSON.stringify([
        { id: 'cur1', type: 'debug', x: 300, y: 400, wires: [] }
    ]),
    addFlowTab: false,
    generateIds: true
})
check('5a: success', r.success === true)
check('5b: stayed on same tab', RED.workspaces.active() === tabBefore)

// 6. Invalid JSON string -> should throw
r = await t('automation/import-flow', { flow: '{not valid json!!!' })
check('6: throws on invalid JSON', r.success === false)

// 7. Empty string -> should do nothing (no error)
r = await t('automation/import-flow', { flow: '' })
check('7: success on empty string', r.success === true)

// 8. Old param name addFlow -> should be ignored
const tabNow = RED.workspaces.active()
r = await t('automation/import-flow', {
    flow: JSON.stringify([
        { id: 'old1', type: 'tab', label: 'Should Not Create' },
        { id: 'old2', type: 'inject', x: 100, y: 100, z: 'old1', wires: [] }
    ]),
    addFlow: true,
    generateIds: true
})
check('8: success', r.success === true)
console.log('>> MANUAL: verify NO new tab created — addFlow is not a recognized param')
```

### Results

All automated checks PASS. Manual checks (undo/redo, tab behavior) confirmed.

@cstns @Steve-Mcl @allthedoll @dimitrieh

andypalmi:
Re: @cstns comment about array support --

The `importFlow` method already handled arrays (they skipped the `typeof === 'string'` branch), but it was implicit. Fixed in latest push:

1. Updated JSDoc to `@param {string|Object[]}` to document both types
2. Added explicit `else if (!Array.isArray(nodesStr))` guard that throws for invalid input types (numbers, objects, etc.)
3. Added unit test for the invalid-type rejection path

Array input was already tested (`should import flow array directly without JSON.parse`), so no test changes needed for that path. The schema `type: ['string', 'array']` is now properly matched by the method.

/cc @cstns @Steve-Mcl @allthedoll @dimitrieh

resources/expertAutomations.js
```diff
@@ -318,8 +340,11 @@ export class ExpertAutomations extends ExpertActionsInterface {
                 e.code = 'NODE_RED'
                 throw e
             }
+        } else if (!Array.isArray(nodesStr)) {
```
Steve-Mcl: like the string version calls `this.redOps.validateFlowString` we should probably add a `this.redOps.validateFlow` method that extracts the functional part of `validateFlowString` into `validateFlow` 

e.g. 

`resources/redOps.js`
```js
    /**
     * Validates if the provided string looks like valid flow json
     * @param {string} flowString the string to validate
     * @returns If valid, returns the node array
     */
    validateFlowString (flowString) {
        return this.validateFlow(JSON.parse(flowString))
    }

    /**
     * Validates if the provided flow looks like valid flow
     * @param {Array<object>} flowArray the flow to validate
     * @returns If valid, returns the node array
     */
    validateFlow (flowArray) {
        if (!Array.isArray(flowArray)) {
            throw new Error(this.RED._('clipboard.import.errors.notArray'))
        }
        for (let i = 0; i < flowArray.length; i++) {
            if (typeof flowArray[i] !== 'object') {
                throw new Error(this.RED._('clipboard.import.errors.itemNotObject', { index: i }))
            }
            if (!Object.hasOwn(flowArray[i], 'id')) {
                throw new Error(this.RED._('clipboard.import.errors.missingId', { index: i }))
            }
            if (!Object.hasOwn(flowArray[i], 'type')) {
                throw new Error(this.RED._('clipboard.import.errors.missingType', { index: i }))
            }
        }
        return flowArray
    }
```


then in this file at line 343:

```js
        } else {
            try {
                newNodes = this.redOps.validateFlow(nodesStr)
            } catch (err) {
                const e = new Error(this.RED._('clipboard.invalidFlow', { message: err.message }))
                e.code = 'NODE_RED'
                throw e
            }
        }
```

resources/expertAutomations.js
```diff
@@ -318,8 +340,11 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+            throw new Error('importFlow expects a JSON string or an array of node objects')
         }
-        this.RED.view.importNodes(newNodes, { generateIds, addFlow, notify })
+        this.RED.view.importNodes(newNodes, { generateIds, addFlow, notify, touchImport: true, applyNodeDefaults: true })
```
Steve-Mcl: If nodes are being imported on to an existing flow tab and that tab is locked, this will not throw - and `notify: false` is not respected by Node-RED (it shows whether `true` or `false`)

Steve-Mcl (commented):
See comments.

#### Detail about "nodes being imported on to an existing flow tab and that tab is locked"

We probably need to check if the target TAB is locked (and throw error if it is) because the RED noftication cannot be inhibited AND no error is thrown.

chrome_xEAs2C3VBT

Steve-Mcl (approved):

cstns (approved):

PR #264, opened 2026-04-15
feat(set-links): add automation/set-links action for link nodes

### LINKED ISSUE #261: feat: add automation/set-links action for link nodes
## Description

Link nodes (`link in`, `link out`, `link call`) do not use the standard `wires` array. Instead, they use a bidirectional `links` property — an array of node IDs on each side.

When connecting `link out` A → `link in` B:
- A's `links` array needs B's ID
- B's `links` array needs A's ID

While `update-node` could technically do this, it requires two separate calls and puts the burden on the caller to manage bidirectional updates correctly. A dedicated `set-links` action would handle this atomically with proper validation.

## Link node types

| Type | `links` connects to | inputs | outputs |
|---|---|---|---|
| `link in` | `link out` node IDs | 0 | 1 |
| `link out` | `link in` node IDs (mode: `link`) | 1 | 0 |
| `link call` | `link in` node IDs (static mode) | 1 | 1 |

## Acceptance criteria

- [ ] `automation/set-links` action with `add`/`remove` modes
- [ ] Validates node existence and compatible types
- [ ] Bidirectional update (both nodes' `links` arrays updated)
- [ ] Duplicate link prevention
- [ ] History/undo support
- [ ] Locked workspace check
- [ ] Unit tests
- [ ] Manual tests in product-assistant.js

andypalmi:
## Summary

- New `automation/set-links` action for managing virtual links between link nodes (`link in`, `link out`, `link call`)
- `link out` <> `link in` connections are **bidirectional** (both nodes' `links` arrays updated)
- `link call` > `link in` connections are **unidirectional** (only `link call` updated)
- Triggers `RED.editor.validateNode` on both nodes after modification to keep NR editor state consistent

## Validation

| Check | Error |
|---|---|
| Self-link | `Cannot link a node to itself` |
| Source not found | `Source node <id> not found` |
| Target not found | `Target node <id> not found` |
| Source not link out/call | `must be a link out or link call node` |
| Target not link in | `must be a link in node` |
| Link out in return mode | `return mode cannot have outbound links` |
| Workspace locked (source) | `Cannot modify links -- workspace <id> is locked` |
| Workspace locked (target) | `Cannot modify links -- workspace <id> is locked` |
| Duplicate link (add) | `Link already exists from <source> to <target>` |
| Missing link (remove) | `Link not found from <source> to <target>` |

## Manual test results

All 13 manual tests passed against a live NR editor:

| Scenario | Result |
|---|---|
| Add link: link out to link in (same tab) | success |
| Remove link: link out to link in | success |
| Add link: link call to link in (unidirectional) | success |
| Remove link: link call to link in | success |
| Self-link | rejected |
| Source node not found | rejected |
| Source is not a link node | rejected |
| Target is not a link in | rejected |
| Link out in return mode | rejected |
| Duplicate link (add same twice) | first succeeds, second rejected |
| Remove non-existent link | rejected |
| Locked workspace | rejected |
| Cross-tab link (link out on tab1, link in on tab2) | success |

Also confirmed: link out can connect to multiple link in nodes (no artificial limit).

## Test plan

- 16 unit tests covering all validation paths, bidirectional/unidirectional behaviour, history/undo, and cross-tab links
- 13 manual console tests against live NR editor
- All 221 unit tests passing

Closes #261

cc @Steve-Mcl -- this addresses the link node handling you flagged during set-wires review. Implemented as a separate action rather than extending set-wires, since link nodes use a `links` property array (not `wires`) with bidirectional semantics.

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -660,6 +660,10 @@ export class ExpertAutomations extends ExpertActionsInterface {
         if (sourceNode.type === 'link out' && sourceNode.mode === 'return') {
             throw new Error(`Source node ${source} is a link out in return mode and cannot have outbound links`)
         }
+        // link call in "dynamic" mode cannot have static links
+        if (sourceNode.type === 'link call' && sourceNode.linkType === 'dynamic') {
+            throw new Error(`Source node ${source} is a link call in dynamic mode and cannot have static links`)
+        }
         // Check workspace locks for both nodes
         if (sourceNode.z && this.RED.workspaces.isLocked(sourceNode.z)) {
             throw new Error(`Cannot modify links — workspace ${sourceNode.z} is locked`)
@@ -680,7 +684,8 @@ export class ExpertAutomations extends ExpertActionsInterface {
             }
             // Record history before mutating
             editHistories.push({ t: 'edit', node: sourceNode, changes: { links: [...sourceLinks] }, changed: wasSourceChanged })
-            sourceNode.links = [...sourceLinks, target]
+            // link call can only have one target — replace instead of append
+            sourceNode.links = sourceNode.type === 'link call' ? [target] : [...sourceLinks, target]
             if (isBidirectional) {
                 editHistories.push({ t: 'edit', node: targetNode, changes: { links: [...targetLinks] }, changed: wasTargetChanged })
                 targetNode.links = [...targetLinks, source]
@@ -707,6 +712,11 @@ export class ExpertAutomations extends ExpertActionsInterface {
         if (this.RED.editor?.validateNode) {
             this.RED.editor.validateNode(sourceNode)
             this.RED.editor.validateNode(targetNode)
+        }
+        // Refresh selection to update virtual link wires on the canvas
+        const selection = this.RED.view.selection()
+        if (selection?.nodes) {
+            this.RED.view.select({ nodes: selection.nodes })
         }
         this.RED.view.redraw()
     }
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -656,6 +656,10 @@ export class ExpertAutomations extends ExpertActionsInterface {
         }
         this.RED.history.push({ t: 'multi', events: editHistories, dirty: wasDirty })
         this.RED.nodes.dirty(true)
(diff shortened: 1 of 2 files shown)


Steve-Mcl (changes requested):
* If the virtual wire is visible (i.e. one of the link nodes are selected), the wire remains visible after being removed.
* If the source or target are selected, the virtual wire is not shown after "add" operation (user has to deselect then reselect the link node)
* When "add" is performed to set the target of a link-call node is should NOT append the array, it should replace the single item in the links array (A link call can only have 1 target)
* If a link call is set to dynamic mode, it should throw an error (the LLM can update the node first)

lastly, can we be consistent with parameter names? The add-wires automation uses `to`/`from` whereas this PR uses  `source`/`target` for param names? (FWIW, I prefer source/target)

Steve-Mcl:
> lastly, can we be consistent with parameter names? The add-wires automation uses `to`/`from` whereas this PR uses `source`/`target` for param names? (FWIW, I prefer source/target)

https://github.com/FlowFuse/nr-assistant/issues/267

andypalmi:
@Steve-Mcl 

> If the virtual wire is visible (i.e. one of the link nodes are selected), the wire remains visible after being removed.
Fixed with last commit [a227c07](https://github.com/FlowFuse/nr-assistant/pull/264/commits/a227c0736cda7df3ff2f42592b18d7e396f64b6d)

> If the source or target are selected, the virtual wire is not shown after "add" operation (user has to deselect then reselect the link node)
Fixed with last commit [a227c07](https://github.com/FlowFuse/nr-assistant/pull/264/commits/a227c0736cda7df3ff2f42592b18d7e396f64b6d)

https://github.com/user-attachments/assets/8af1be6a-5575-4871-90b1-e949b0329851

> When "add" is performed to set the target of a link-call node is should NOT append the array, it should replace the single item in the links array (A link call can only have 1 target)
Returns `"Link already exists from link_call_id to link_in_id"`

> If a link call is set to dynamic mode, it should throw an error (the LLM can update the node first)
Returns `"Source node link_call_id is a link call in dynamic mode and cannot have static links"`

Steve-Mcl (approved):

PR #271, opened 2026-04-16
feat(update-node): add line-based partial edits via patches

### LINKED ISSUE #270: Add line-based partial edits to update-node action
## Summary

The `update-node` action currently supports full property replacement only (via `Object.assign`). This is insufficient for editing multiline string properties like `func` (function node code), `template`, or `info` where only specific lines need to change.

## Proposal

Extend the `update-node` action with an optional `patches` parameter that supports line-based partial edits on string properties.

Each patch specifies:
- `property` — the node property to edit
- `from` / `to` — 1-indexed inclusive line range to replace
- `content` — replacement text (`\n` for multiple lines, empty string to delete)

Insert mode: when `from = to + 1`, content is inserted without removing lines (supports append, prepend, and insert-between).

`patches` and `properties` can coexist in the same call (e.g., rename node + patch code). At least one must be provided.

## Validation

- Line numbers must be positive integers within bounds
- Overlapping patches on the same property are rejected
- Target property must be a string
- Multiple patches applied bottom-up to avoid line-number drift

andypalmi:
## Summary

- Extend the `update-node` action with an optional `patches` parameter for line-based partial edits on string properties (`func`, `template`, `info`, etc.)
- `patches` and `properties` can coexist in the same call (e.g., rename a node and patch its code at once)
- All changes are recorded in history for undo/redo support

Closes #270

## Patch format

Each patch is an object with `property`, `from`, `to`, and `content`:

```json
{
  "id": "node-id",
  "patches": [
    { "property": "func", "from": 2, "to": 3, "content": "replacement lines" }
  ]
}
```

Lines are 1-indexed and inclusive.

### Replace lines

Replace one or more lines with new content. The replacement can have fewer, equal, or more lines than the range.

```json
{ "property": "func", "from": 2, "to": 2, "content": "const b = 10;" }
{ "property": "func", "from": 2, "to": 4, "content": "single replacement" }
{ "property": "func", "from": 1, "to": 1, "content": "line1\nline2\nline3" }
```

### Delete lines

Set `content` to an empty string to remove lines.

```json
{ "property": "func", "from": 3, "to": 4, "content": "" }
```

### Insert lines (append, prepend, insert between)

Set `from = to + 1` to insert without removing any lines:

```json
{ "property": "func", "from": 6, "to": 5, "content": "appended after line 5" }
{ "property": "func", "from": 1, "to": 0, "content": "prepended before line 1" }
{ "property": "func", "from": 4, "to": 3, "content": "inserted between lines 3 and 4" }
```

### Multiple patches in one call

Non-overlapping patches on the same property are applied bottom-up to avoid line-number drift:

```json
{
  "patches": [
    { "property": "func", "from": 1, "to": 1, "content": "// top" },
    { "property": "func", "from": 5, "to": 5, "content": "// bottom" }
  ]
}
```

### Replace + insert combo

Edit a line and insert new lines adjacent to it in the same call. Works at both ends of the property:

```json
{
  "patches": [
    { "property": "func", "from": 5, "to": 5, "content": "return [msg, null];" },
    { "property": "func", "from": 6, "to": 5, "content": "// new line after last" }
  ]
}
```

```json
{
  "patches": [
    { "property": "func", "from": 1, "to": 0, "content": "// header" },
    { "property": "func", "from": 1, "to": 1, "content": "const a = 99;" }
  ]
}
```

### Mixed with properties

Patch code and update other node fields in the same call:

```json
{
  "id": "node-id",
  "properties": { "name": "Patched Node" },
  "patches": [
    { "property": "func", "from": 1, "to": 1, "content": "// patched first line" }
  ]
}
```

### Patches on different properties

A single call can patch multiple string properties independently:

```json
{
  "patches": [
    { "property": "func", "from": 1, "to": 1, "content": "new code" },
    { "property": "template", "from": 2, "to": 2, "content": "<h2>Updated</h2>" }
  ]
}
```

## Validation

- `from` and `to` must be integers within bounds of the target property's line count
- Overlapping patches on the same property are rejected
- Target property must be a string
- At least one of `properties` or `patches` must be provided
- Empty `properties: {}` is rejected

## Test plan

- Replace: single line, range with fewer/more lines
- Delete: lines removed via empty content
- Insert: append after last line, prepend before first line, insert between lines
- Combo: replace last line + append, prepend + replace first line
- Multi-patch: two non-overlapping patches applied bottom-up correctly
- Cross-property: patches on func and template independently
- Mixed: patches + properties together, history records originals for both
- Undo: history captures original property values before mutation
- Single-line property edge case
- Error paths: overlapping patches, out-of-range lines, non-string property, non-integer from/to, missing params, empty properties, node not found


Steve-Mcl:
Not keen on `from = to + 1` to mean "insert" (I dislike magic numbers and not-so-obvious overloading)
Also, not keen on `to`/`from` (feels ambiguous)
This feels like it needs an operand/operator field.

Additionally, some node properties are nested (e.g. in a `change` node, the 3rd row setup to set  `msg,payload` to a JSONata value would be property `rules[2].to`) this implementation is shallow and would probably break external code if changed later to support property paths.

 We should probably look at something like JSON PATH RFC (https://datatracker.ietf.org/doc/html/rfc6902) for inspiration?

Perhaps something like is a bit more flexible?:
```js
patches: {
    type: 'array',
    description: 'Line-based partial edits for string properties. All line numbers reference the original content before any patches are applied.',
    items: {
        type: 'object',
        properties: {
            property: { type: 'string', description: 'Top-level node property name (used for undo snapshot)' },
            path: { type: 'string', description: 'Optional dot-separated path within the property value to reach the target string. Numeric segments index into arrays (e.g. "0.to" for rules[0].to). Omit if the property itself is the target string.' },
            op: {
                type: 'string',
                enum: ['replace', 'insert', 'delete'],
                description: 'replace: replace lines start..end. insert: insert before start. delete: remove lines start..end.'
            },
            start: { type: 'number', description: 'Start line (1-indexed, inclusive)' },
            end: { type: 'number', description: 'End line (1-indexed, inclusive). Required for replace/delete.' },
            content: { type: 'string', description: 'Text to insert or replace with (\\n for multiple lines). Required for replace/insert.' }
        },
        required: ['property', 'op', 'start']
    }
}
```
e.g.  Patch a change node's JSONata expression (nested):
`{ "op": "replace", "property": "rules", "path": "0.to", "start": 1, "end": 1, "content": "$sum(payload.items.price) * 1.2" }`

e.g. Patch a function node's code (top-level, no path needed — same as today):
`{ "op": "replace", "property": "func", "start": 5, "end": 7, "content": "const result = payload * 2;\nreturn msg;" }`


NOTE:

The above is a proposal. I am more than happy to see a single propertyPath instead of separate fields (e.g. `"property": "rules.0.to"` or `"property": "rules[0].to"` or `"path": ["rules", 0, "to"]` - whichever is best to support regular dot notation, object property by name notation, and Array indexes without confusing one for the other)

One other thing to note, it will probably need support for a `createIfMissing` capability (defaulted to `false` is my gut feel)

andypalmi:
Updated to adopt the suggested schema. Changes pushed.

**Schema**: Replaced `from`/`to`/`content` with explicit `op` (`replace`, `insert`, `delete`), `start`, `end`, and `content` fields. `end` is required for replace/delete, `content` is required for replace/insert. `required` is now `['property', 'op', 'start']`.

**Nested property paths**: Added optional `path` field for dot-separated navigation into nested structures. Numeric segments index into arrays. For example, to patch a change node's JSONata expression:
```json
{ "op": "replace", "property": "rules", "path": "0.to", "start": 1, "end": 1, "content": "$sum(payload.items.price) * 1.2" }
```
When `path` is omitted, the property itself is the target string (same as before, just with the new field names).

For undo, the entire top-level property is deep-cloned when a nested path is used, so `Ctrl+Z` restores the full original value.

**Sidebar refresh**: Added `RED.sidebar.info.refresh()` call after node updates so the info panel reflects changes immediately.

**Validation per op**:
- `replace`: requires `end` and `content`, `start <= end`, `end <= lineCount`
- `insert`: requires `content`, `start <= lineCount + 1`
- `delete`: requires `end`, `start <= end`, `end <= lineCount`
- Overlapping ranges checked across replace/delete ops only (inserts don't consume lines)

All 282 tests pass (30 patch-specific tests covering replace, insert, delete, nested paths, combo ops, sidebar refresh, and error paths).

RE `createIfMissing` -- not implemented yet, can add in a follow-up if needed.

andypalmi:
Added auto-detect for line separators. Some Node-RED properties use tab characters as line separators instead of newlines (e.g. inject node JSONata expressions stored via `props[0].v`). The patch system now detects the separator used in the target string and preserves it when applying edits. Patch content always uses `\n` — conversion happens transparently.

Also added unit tests for tab separator handling (replace, insert, delete, mixed tabs+newlines fallback to newlines).

andypalmi:
@Steve-Mcl  When you have time, could you please review it again and merge it ? Thank you

resources/expertAutomations.js
```diff
@@ -133,9 +133,29 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
-                    properties: { type: 'object', description: 'Key-value pairs to merge into the node object' }
+                    properties: { type: 'object', description: 'Key-value pairs to merge into the node object' },
+                    patches: {
+                        type: 'array',
+                        description: 'Line-based partial edits for string properties. All line numbers reference the original content before any patches are applied.',
+                        items: {
+                            type: 'object',
+                            properties: {
+                                property: { type: 'string', description: 'Top-level node property name (used for undo snapshot)' },
+                                path: { type: 'string', description: 'Optional dot-separated path within the property value to reach the target string. Numeric segments index into arrays (e.g. "0.to" for rules[0].to). Omit if the property itself is the target string.' },
```
Steve-Mcl: On reflection, having been working on the flow gen stuff over last few days, I think `property` and `path` should be a singular `property`.

something like
```js
property: { type: 'string', description: 'The dot-separated property path of the node to update. Use numeric segments for index arrays (e.g. use `rules.0.to` to update `rules[0].to`).' }
```

Obviously changing this will require changes to the rest of the PR

resources/expertAutomations.js
```diff
@@ -452,6 +497,180 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+            const closeAll = () => {
+                this.RED.tray.close()
+                setTimeout(() => {
+                    if (this.RED.view.state() !== this.RED.state?.DEFAULT) {
+                        closeAll()
+                    }
+                }, 300)
+            }
+            closeAll()
+        }
```
Steve-Mcl: Couple of things here:

1. calling tray close does not call the individual nodes own `oneditcancel` unfortunately. This, in some cases, can have a side effects e.g. closing the function editor without it calling its own cleanup routines in `oneditcancel` results in  : Image
2. The use of `setTimeout` means we cannot await this action (not possible to sync-lock it for sequential subsequent actions)

Lets extract this to its own callable action (e.g. something like automation/close_editor_tray) then call it from this tool

Also, instead of setTimeout (which cannot be awaited), we should just loop (probably with a MAX limit) something like

```js
async closeEditorTray() {
    const MAX = 10
    let count = 0
    while(count++ < MAX) {
       $('.red-ui-tray-toolbar button#node-dialog-cancel').trigger('click') // this is idempotent so ok to call regardless of state
        await new Promise(r => setTimeout(r, 300))
        if (this.RED.view.state() === this.RED.state?.DEFAULT) break
    }
    return this.RED.view.state() === this.RED.state?.DEFAULT
}
```

^^^ UNTESTED ^^^

Steve-Mcl (changes requested):
There are some issues with the close routine (see comments)

Also not super keen on this being mixed in (overloading) the `update_node` action but I wont block for that.

resources/expertAutomations.js
Steve-Mcl: NOTE on the above. 

The selector `'.red-ui-tray-toolbar button#node-dialog-cancel'` looks a bit verbose - it is deliberate. For those times when a node author has also called their own button `#node-dialog-cancel` (we may even tighten it up more, but this is specific enough for 1st iteration)

andypalmi: Hi @Steve-Mcl , I implemented your suggestions from above in commit [28e3189](https://github.com/FlowFuse/nr-assistant/pull/271/commits/28e3189b11950daddabe3f7a487d75b67b0811df)

andypalmi:
Hi @Steve-Mcl can you review this one and get it merged? This looks ready to me after the tests I have been doing.

Thank you

Steve-Mcl (commented):
Not a full review, just a comment for @cstns - I the only gotcha I see here is if the edit is for a function node code, the final code will not actually be validated. This is more of a Node-RED issue tho (RED.editor.validateNode wont instantiate a monaco instance and validate the funcion code)

Will let you decide how we proceed (may be a follow up on assistant or on Node-RED

andypalmi:
Thank you @Steve-Mcl , I will see how to tackle the validation of a function node in another issue or PR to separate concerns. Would that be alright?

resources/expertAutomations.js
```diff
@@ -133,9 +135,28 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+                                op: {
+                                    type: 'string',
+                                    enum: ['replace', 'insert', 'delete'],
+                                    description: 'replace: replace lines start..end. insert: insert content before line start. delete: remove lines start..end.'
+                                },
+                                start: { type: 'number', description: 'Start line (1-indexed, inclusive)' },
+                                end: { type: 'number', description: 'End line (1-indexed, inclusive). Required for replace and delete.' },
+                                content: { type: 'string', description: 'Text to insert or replace with (\\n for multiple lines). Required for replace and insert.' }
+                            },
+                            required: ['property', 'op', 'start']
```
cstns, line 155: shouldn't end and content also be required?

resources/expertAutomations.js
```diff
@@ -287,6 +308,9 @@ export class ExpertAutomations extends ExpertActionsInterface {
                 },
                 required: ['flow']
             }
+        },
```
cstns, line 311: editor tray changes were not mentioned in the pr

resources/expertAutomations.js
```diff
@@ -824,7 +1045,7 @@ export class ExpertAutomations extends ExpertActionsInterface {
             break
 
         case UPDATE_NODE:
-            this.updateNode(params.id, params.properties)
+            await this.updateNode(params.id, params.properties, params.patches)
```
cstns, line 1048: should we not send the updated node back to the agent to save another round trip later on to interrogate if the changes were applied successfully?

resources/expertAutomations.js
```diff
@@ -912,6 +1133,11 @@ export class ExpertAutomations extends ExpertActionsInterface {
             this.importFlow(params.flow, { addFlow: params.addFlowTab, generateIds: params.generateIds ?? true })
             result.success = true
             break
+
+        case CLOSE_EDITOR_TRAY:
+            result.closed = await this.closeEditorTray()
+            result.success = true
```
cstns, line 1139: shouldn't success depend on the fact that the editor tray was open and now it's closed?

resources/expertAutomations.js
```diff
@@ -452,6 +501,178 @@ export class ExpertAutomations extends ExpertActionsInterface {
             this.RED.editor.validateNode(node)
```
cstns, line 501: as per @steve's comment
> Not a full review, just a comment for @cstns - I the only gotcha I see here is if the edit is for a function node code, the final code will not actually be validated. This is more of a Node-RED issue tho (RED.editor.validateNode wont instantiate a monaco instance and validate the funcion code)

validating this through NR's monaco editor might get tricky, even more so that the monaco instance model needs to be instantiated in the UI to semantically validate the patched fn code.

I align with steve, it's more of a NR issue than an assistant one, and we can't/shouldn't pin the release of this feature due to it.

Just typing out loud here, as a quick and dirty workaround we could (in case we're patching a fn node): 
1. patch the fn node
2. issue an action to open the updated node in the ui
2. post/get through http to the UI to get the result from the instantiated monaco editor syntax / semantic validation
3. throw in case of failure response / continue if valid

resources/expertAutomations.js
andypalmi: I don't think we should as the information sent back might be big, as it should instead be a behaviour of the agent to decide how to proceed but LMK how you think we should proceed. Claude Code does not send back the full payload after editing but only whether it succeeded or not for example

@Steve-Mcl any thoughts?

resources/expertAutomations.js
andypalmi: IMO no, because the tray:
* if open, gets closed
* if closed, remains closed.

It's just a way to improve UX since we cannot do live changes on the editor yet.

resources/expertAutomations.js
andypalmi: Should we handle it in this PR or open a new issue?

Steve-Mcl: new PR

resources/expertAutomations.js
andypalmi: No — end and content are conditionally required depending on op:
- replace → needs both end and content
- delete → needs end, no content
- insert → needs content, no end

resources/expertAutomations.js
Steve-Mcl: its fine idempotent

resources/expertAutomations.js
Steve-Mcl: follow up (either in core or assistant)

Steve-Mcl (approved):

PR #286, opened 2026-04-24
feat(actions): return contextual payloads from all write actions

### LINKED ISSUE #285: Return contextual payload from all automation actions
## Problem

Most write actions in `expertAutomations.js` only return `{ success: true }` with no contextual data. The caller never sees the actual result of what it asked for -- for example, after creating a node, it doesn't get back the created node object with its confirmed ID.

This causes issues with ID tracking across sequential operations and provides no feedback loop to course-correct on failures.

## Proposed changes

Enrich the response payload for every action that currently returns only `success: true`:

| Action | Current return | Proposed return |
|--------|---------------|-----------------|
| `UPDATE_NODE` | `success: true` | `node: {formatted updated node}` |
| `SHOW_WORKSPACE` | `success: true` | `workspace: {id, label}` |
| `ADD_TAB` | `success: true` | `tab: {formatted tab node}` |
| `REMOVE_TAB` | `success: true` | `removedId` |
| `ADD_NODES` | `success: true` | `nodes: [formatted created nodes]` |
| `REMOVE_NODES` | `success: true` | `removedIds: [ids]` |
| `SET_WIRES` | `success: true` | `wires: {echo back connection params}` |
| `SET_LINKS` | `success: true` | `links: {echo back link params}` |
| `IMPORT_FLOW` | `success: true` | `nodes: [formatted imported nodes]` |

Actions that already return contextual data (`SELECT_NODES`, `GET_NODES`, `EDIT_NODE`, `SEARCH`, `ADD_FLOW_TAB`, `GET_FLOW`, `LIST_WORKSPACES`, `CLOSE_EDITOR_TRAY`) and UI-only actions (`CLOSE_SEARCH`, `CLOSE_TYPE_SEARCH`, `CLOSE_ACTION_LIST`) need no changes.

## Context

Builds on top of #270.

andypalmi:
## Summary

- Enrich response payloads for all write actions that previously only returned `{ success: true }`
- `ADD_NODES` now returns the formatted created nodes
- `UPDATE_NODE` now returns the formatted updated node
- `ADD_TAB` now returns the formatted tab node
- `REMOVE_TAB` / `REMOVE_NODES` now echo back the removed ID(s)
- `SET_WIRES` / `SET_LINKS` now echo back the connection params
- `IMPORT_FLOW` now returns the formatted imported nodes
- `SHOW_WORKSPACE` now returns the workspace id and label
- `importFlow()` and `addTab()` helper methods updated to return their results

## Test plan

- [x] All 299 existing tests pass
- [x] Updated test assertions to verify new return properties on each affected action
- [x] Verified `addNodes` test handles pre-import existence check vs post-import lookup correctly
- [x] Verified `importFlow` handles the returned array from `RED.view.importNodes`

Closes #285

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -1062,7 +1062,7 @@ export class ExpertAutomations extends ExpertActionsInterface {
         case SHOW_WORKSPACE: {
             this.showWorkspace(params.id)
             const ws = this.RED.nodes.workspace(params.id)
-            result.data = ws ? { id: ws.id, label: ws.label } : null
+            result.data = ws ? { id: ws.id, label: typeof ws.label === 'function' ? ws.label() : ws.label } : null
             result.success = true
         }
             break
@@ -1085,7 +1085,7 @@ export class ExpertAutomations extends ExpertActionsInterface {
         case LIST_WORKSPACES: {
             const workspaceIds = this.RED.nodes.getWorkspaceOrder()
             const tabs = workspaceIds.map(id => this.RED.nodes.workspace(id))
-            const sanitized = tabs.map(t => ({ id: t.id, label: t.label, disabled: t.disabled, info: t.info, locked: t.locked, contentsChanged: t.contentsChanged }))
+            const sanitized = tabs.map(t => ({ id: t.id, label: typeof t.label === 'function' ? t.label() : t.label, disabled: t.disabled, info: t.info, locked: t.locked, contentsChanged: t.contentsChanged }))
             const selectedWorkspaces = this.RED.workspaces.selection() || []
             sanitized.forEach(t => {
                 t.hidden = this.RED.workspaces.isHidden(t.id)
@@ -1178,7 +1178,7 @@ export class ExpertAutomations extends ExpertActionsInterface {
         const s = { id: node.id }
         if (node.type !== undefined) s.type = node.type
         if (node.name !== undefined) s.name = node.name
-        if (node.label !== undefined) s.label = node.label
+        if (node.label !== undefined && typeof node.label !== 'function') s.label = node.label
         if (node.x !== undefined) s.x = node.x
         if (node.y !== undefined) s.y = node.y
         if (node.z !== undefined) s.z = node.z
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -1022,7 +1022,7 @@ export class ExpertAutomations extends ExpertActionsInterface {
             break
         case EDIT_NODE: {
             const selectedNode = this.editNode(params.id || '')
-            result.node = this._formatNodes([selectedNode], false)[0] || null
+            result.node = this._formatNodes([selectedNode], params.options?.includeModuleConfig)[0] || null
             result.success = true
         }
             break
@@ -1032,7 +1032,7 @@ export class ExpertAutomations extends ExpertActionsInterface {
                 result.results = []
                 for (let index = 0; index < searchResults.length; index++) {
                     const searchResult = searchResults[index]
-                    searchResult.node = this._formatNodes([searchResult.node], false)[0] || null
+                    searchResult.node = this._formatNodes([searchResult.node], params.options?.includeModuleConfig)[0] || null
                     result.results.push(searchResult)
                 }
             }
@@ -1041,14 +1041,14 @@ export class ExpertAutomations extends ExpertActionsInterface {
             break
         case ADD_FLOW_TAB: {
             const newFlowTab = await this.addFlowTab(params?.title || undefined)
-            result.tab = this._formatNodes([newFlowTab], false)[0] || null
+            result.tab = this._formatNodes([newFlowTab], params.options?.includeModuleConfig)[0] || null
             result.success = true
         }
             break
 
         case UPDATE_NODE:
             await this.updateNode(params.id, params.properties, params.patches)
-            result.node = this._formatNodes([this.RED.nodes.node(params.id)], false)[0] || null
+            result.node = this._formatNodes([this.RED.nodes.node(params.id)], params.options?.includeModuleConfig)[0] || null
             result.success = true
             break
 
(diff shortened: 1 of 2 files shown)


Steve-Mcl:
My only concern with returning results to agent is "are we sending the response to LLM"? If yes, we are gonna hit context issues very quickly.

We may need to be smart about what hits the LLM after an MCP tool call. Instead of sending a crap ton of JSON, we may need to send formatted responses like
- "Created node id:abcdef123456, z:fedcba654321, type:function, x:110, y:200" 
- "Created flow 1) node id:abcdef123456, z:fedcba654321, type:mqtt-out, x:200, y:200, broker:fedcba9999999, wires:[]. 2) config node id:fedcba9999999, type:mqtt-broker." 
- "Updated node id:abcdef123456, x: old=100, new=400, name: old='old name' new='new name'"

andypalmi:
@Steve-Mcl I don't agree with the context issue for now and in any case we could improve the summarisation logic later. 
In my most recent testings, I never got those issues (considering the 200k tokens context window of all latest models) but I could be wrong. All other coding agents get the full payload from tools, so we might need to start with the full payload and later iterate to reduce the context sent back to the agent if needed.

Steve-Mcl:
My preference is the reverse. If we send everything to LLM, we burn tokens, slow it to a crawl and we may even struggle to reduce this at a later time.

My recommendation is we only send a diff to LLM.

e.g. import_flows sends 100kb of JSON to import. It gets imported and its perfect (no diff from requested import to final flow) - what is the point of sending 100Kb of flow back to LLM just to say "that worked"

e.g. update_node wants to set `change` node `rule.0.to` value to a 1kb JSON value. It succeeds. Then we send the full node (including all fields and rules in the node) back to LLM to say "it worked" - whats the point!

If however we had a call that tried to set `made_up_prop` to `123` and a diff reveals this property was not set, then reporting back something explicitly calling that out, its more useful, less verbose, more impactful.

Just like when a schema validation fails and we send the validation results (e.g. "func must be a string"), we should probably consider something similar.

Anyhow, this is just my opinion. Happy to be overruled.

resources/expertAutomations.js
```diff
@@ -422,8 +422,9 @@ export class ExpertAutomations extends ExpertActionsInterface {
         if (!addFlow && this.RED.workspaces.isLocked()) {
             throw new Error('Cannot import into a locked workspace')
         }
-        this.RED.view.importNodes(newNodes, { generateIds, addFlow, touchImport: true, applyNodeDefaults: true })
+        const imported = this.RED.view.importNodes(newNodes, { generateIds, addFlow, touchImport: true, applyNodeDefaults: true })
```
cstns: importNodes seems to be throwing some errors, wouldn't it be wise to catch those errors if they occur?

see @ `packages/node_modules/@node-red/editor-client/src/js/ui/view.js:6181`

resources/expertAutomations.js
```diff
@@ -1039,19 +1041,23 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
             await this.updateNode(params.id, params.properties, params.patches)
+            result.node = this._formatNodes([this.RED.nodes.node(params.id)], params.options?.includeModuleConfig)[0] || null
             result.success = true
             break
 
-        case SHOW_WORKSPACE:
+        case SHOW_WORKSPACE: {
             this.showWorkspace(params.id)
+            const ws = this.RED.nodes.workspace(params.id)
+            result.workspace = ws ? { id: ws.id, label: ws.label } : null
```
cstns: why not return the plain workspace if present?

resources/expertAutomations.js
```diff
@@ -1099,39 +1105,56 @@ export class ExpertAutomations extends ExpertActionsInterface {
             result.success = true
             break
 
-        case ADD_TAB:
-            this.addTab(params)
+        case ADD_TAB: {
+            const newTab = this.addTab(params)
+            result.tab = this._formatNodes([newTab], params.options?.includeModuleConfig)[0] || null
```
cstns: does returning each type of response action on it's own unique key based on the action help the agent in any way? why not standardize the response payload under a generic key? I feel like this approach is adding a lot of unnecessary complexity

resources/expertAutomations.js
```diff
@@ -1099,39 +1105,56 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
             result.success = true
+        }
             break
 
-        case REMOVE_TAB:
+        case REMOVE_TAB: {
+            const ws = this.RED.nodes.workspace(params.id)
+            const tab = ws ? this._formatNodes([ws], params.options?.includeModuleConfig)[0] : null
```
cstns: why return only the first element of the formatted nodes?

resources/expertAutomations.js
```diff
@@ -1039,19 +1048,23 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
             result.success = true
             break
 
-        case SHOW_WORKSPACE:
+        case SHOW_WORKSPACE: {
             this.showWorkspace(params.id)
+            const ws = this.RED.nodes.workspace(params.id)
+            result.data = ws ? { id: ws.id, label: ws.label } : null
             result.success = true
+        }
```
Steve-Mcl, line 1067: Can we be consistent and provide what `LIST_WORKSPACES` returns - e.g. include the locked and active flags etc

Either refactor & reuse the code that LIST_WORKSPACES has (PREFFERED) or duplicate what it does in SHOW_WORKSPACE (YUK)

resources/expertAutomations.js
Steve-Mcl: I think the question is, is there any value in returning the tab that was just removed!?

Perhaps we should return a fresh list of tabs (in same format as LIST_WORKSPACES) so that the LLM can determine /understand what tabs remain.

Remember, LIST_WORKSPACES includes active/locked/disabled flags so it will be useful data to the LLM to know current state after the removal of a tab.

resources/expertAutomations.js
andypalmi: See https://github.com/FlowFuse/nr-assistant/pull/286/changes/e03eef2164553da12b89c5d86809daa9784e7e95

resources/expertAutomations.js
andypalmi: See https://github.com/FlowFuse/nr-assistant/pull/286/changes/e03eef2164553da12b89c5d86809daa9784e7e95

resources/expertAutomations.js
andypalmi: See https://github.com/FlowFuse/nr-assistant/pull/286/changes/8f0d3f0519143718f2576d32ad8f0df806e4c98e

Steve-Mcl (approved):

PR #288, opened 2026-04-24
feat(actions): return node validation results from create/update actions

### LINKED ISSUE #287: Return node validation results from create/update actions
## Problem

When nodes are created or updated via automation actions, the caller gets no feedback on whether the resulting node is valid. Invalid property values (wrong types, missing required config refs, bad enum values) pass silently with just `success: true`.

`RED.editor.validateNode()` is already called during `updateNode()` but its result is discarded. After `addNodes()` and `importFlow()`, no validation runs at all.

## Proposed changes

Return validation results from the three actions that create or mutate nodes:

- `UPDATE_NODE`: capture the result of `RED.editor.validateNode()` and include it in the action response
- `ADD_NODES`: after import, run validation on each created node and include results
- `IMPORT_FLOW`: after import, run validation on each imported node and include results

Validation results should be returned alongside the existing success response (not block the action), so the caller gets a feedback loop to detect and correct invalid configurations.

## Context

Builds on top of #285 / #286.

andypalmi:
## Summary

- After `UPDATE_NODE`, `ADD_NODES`, and `IMPORT_FLOW` actions, return a `validation` field with the node's validity state
- `UPDATE_NODE` returns `validation: { valid, validationErrors? }` for the single updated node
- `ADD_NODES` and `IMPORT_FLOW` return `validation: [{ id, valid, validationErrors? }]` containing only invalid nodes (empty array if all valid)
- Added `_getNodeValidation()` helper with documentation on how Node-RED's `RED.editor.validateNode()` works internally

### How Node-RED validates nodes

`RED.editor.validateNode(node)` is a side-effect function that mutates the node object directly:
- Sets `node.valid` (boolean)
- Sets `node.validationErrors` (string array -- property names or custom error messages)
- Checks each property against `node._def.defaults` validators: required fields, custom validators, config node refs, credentials
- For subflow instances, cascades into the subflow definition and all internal nodes

The `_getNodeValidation()` helper reads these properties back into a compact object for the action response.

## Test plan

- [x] All 302 tests pass (3 new tests added)
- [x] Verified UPDATE_NODE returns `validation: { valid: true }` on success and `{ valid: false, validationErrors: ['prop'] }` on failure
- [x] Verified ADD_NODES returns empty validation array when all nodes are valid, array with invalid entries when not
- [x] Verified IMPORT_FLOW same pattern as ADD_NODES
- [x] Verified `validationErrors` uses array type matching Node-RED's actual implementation

Closes #287

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -1046,10 +1046,13 @@ export class ExpertAutomations extends ExpertActionsInterface {
         }
             break
 
-        case UPDATE_NODE:
+        case UPDATE_NODE: {
             await this.updateNode(params.id, params.properties, params.patches)
-            result.node = this._formatNodes([this.RED.nodes.node(params.id)], params.options?.includeModuleConfig)[0] || null
+            const updatedNode = this.RED.nodes.node(params.id)
+            result.node = this._formatNodes([updatedNode], params.options?.includeModuleConfig)[0] || null
+            result.validation = this._getNodeValidation(updatedNode)
             result.success = true
+        }
             break
 
         case SHOW_WORKSPACE: {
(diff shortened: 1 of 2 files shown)


andypalmi:
@Steve-Mcl @cstns This is pretty straightforward but let me know if you find any issues or have any comments.

Steve-Mcl (approved):

PR #291, opened 2026-04-27
feat(actions): add automation/get-node-types action

### LINKED ISSUE #290: feat(actions): add automation/get-node-type action for runtime node type discovery
## Summary

Add a new `automation/get-node-type` read action that queries `RED.nodes.getType(type)` at runtime and returns the node definition's defaults, label, category, color, and port counts.

## Motivation

`RED.nodes.getType(type)` is available at runtime for any installed node type, but there is currently no action that exposes this data. Having it as an action allows callers to discover property names, default values, and node metadata for any installed node — including community packages — without relying on a pre-built schema.

## Action spec

**Name:** `automation/get-node-type`

**Params:**
- `type` (string, required) — node type identifier

**Returns:**
- `type` — the node type string
- `defaults` — node defaults object (`{ propName: { value, ... }, ... }`)
- `label` — display label (string or serialized function)
- `category` — palette category
- `color` — background color
- `inputs` — number of input ports
- `outputs` — number of output ports

Returns `success: false` with an error if the type is not installed in this Node-RED instance.

andypalmi:
## Summary

- Add `automation/get-node-type` action that calls `RED.nodes.getType(type)` at runtime
- Returns `defaults`, `label`, `category`, `color`, `inputs`, `outputs` for any installed node type
- Returns `success: false` with a descriptive error if the type is not installed
- Add action to the supported actions list and update tests

## Test plan

- Existing test suite passes (302 tests)
- `should have supported actions` test updated to include `automation/get-node-type`

Closes #290

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -1195,18 +1195,10 @@ export class ExpertAutomations extends ExpertActionsInterface {
                 return
             }
             const rawDefaults = def.defaults || {}
-            result.type = params.type
-            result.defaults = Object.fromEntries(
-                Object.entries(rawDefaults).map(([key, propDef]) => {
-                    if (propDef && typeof propDef === 'object') {
-                        const safe = Object.fromEntries(
-                            Object.entries(propDef).filter(([, v]) => typeof v !== 'function')
-                        )
-                        return [key, safe]
-                    }
-                    return [key, propDef]
-                })
-            )
+            result.nodeType = params.type
+            result.defaults = JSON.parse(JSON.stringify(rawDefaults, (key, value) =>
+                typeof value === 'function' ? value.toString() : value
+            ))
             result.label = typeof def.label === 'function' ? def.label.toString() : (def.label || params.type)
             result.category = def.category || null
             result.color = def.color || null
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -21,7 +21,7 @@ const SET_WIRES = 'automation/set-wires'
 const SET_LINKS = 'automation/set-links'
 const IMPORT_FLOW = 'automation/import-flow'
 const CLOSE_EDITOR_TRAY = 'automation/close-editor-tray'
-const GET_NODE_TYPE = 'automation/get-node-type'
+const GET_NODE_TYPES = 'automation/get-node-types'
 
 /**
  * @typedef {SELECT_NODES
@@ -44,7 +44,7 @@ const GET_NODE_TYPE = 'automation/get-node-type'
  *   |SET_LINKS
  *   |IMPORT_FLOW
  *   |CLOSE_EDITOR_TRAY
- *   |GET_NODE_TYPE} ExpertAutomationsActionsEnum
+ *   |GET_NODE_TYPES} ExpertAutomationsActionsEnum
  */
 
 export class ExpertAutomations extends ExpertActionsInterface {
@@ -314,14 +314,16 @@ export class ExpertAutomations extends ExpertActionsInterface {
         [CLOSE_EDITOR_TRAY]: {
             params: null
         },
-        [GET_NODE_TYPE]: {
+        [GET_NODE_TYPES]: {
             params: {
                 type: 'object',
-                required: ['type'],
+                required: ['types'],
                 properties: {
-                    type: {
-                        type: 'string',
-                        description: 'Node type identifier to look up (e.g. "inject", "function", "worldmap")'
+                    types: {
+                        type: 'array',
+                        items: { type: 'string' },
+                        minItems: 1,
+                        description: 'One or more node type identifiers to look up (e.g. ["inject", "function", "ui-text"])'
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


Steve-Mcl (commented):
is this pr merging into the wrong branch (I can see update_node mods!)

also, is get-node-type (singular) not a bit restrictive considering how the agent is put together?

Perhaps this should support an array of types like get-node-types ['function','ui-text'] (which can then be called with one element to support singular requests)

Lastly, should this not return a JSON schema in alignment with new schema direction (i.e. so that when Node-RED has built in schema support we simply return that instead of a made-up format for assistant for GET_NODE_TYPES)?

andypalmi:
It's merging in the right branch, I made all PRs sequential

andypalmi:
> also, is get-node-type (singular) not a bit restrictive considering how the agent is put together?
> Perhaps this should support an array of types like get-node-types ['function','ui-text'] (which can then be called with one element to support singular requests)

IMO having it singular or array doesn't make that much of a difference but please let me know how you guys would like to proceed. To have it standardised

> Lastly, should this not return a JSON schema in alignment with new schema direction (i.e. so that when Node-RED has built in schema support we simply return that instead of a made-up format for assistant for GET_NODE_TYPES)?

We have no official schema yet, do we want a version 0 for this too or wait?

@Steve-Mcl @cstns

Steve-Mcl:
> It's merging in the right branch, I made all PRs sequential

If it were a branch thing AND the correct branch was selected, the changes from the child branch would not be highlighted.

Look at the green sections in this PR - there are changes in this PR to both [updateNode](https://github.com/FlowFuse/nr-assistant/pull/291/changes#diff-b89bcd54930b5abba7a5ccddc78ddc849b36e62d572f817a4aa31d49900cf09dR485) signature and the [UPDATE_NODE](https://github.com/FlowFuse/nr-assistant/pull/291/changes#diff-b89bcd54930b5abba7a5ccddc78ddc849b36e62d572f817a4aa31d49900cf09dR1097-R1122) signature

Steve-Mcl:
> > Lastly, should this not return a JSON schema in alignment with new schema direction (i.e. so that when Node-RED has built in schema support we simply return that instead of a made-up format for assistant for GET_NODE_TYPES)?
> 
> We have no official schema yet, do we want a version 0 for this too or wait?

What i worry about here is lock-in. Once we ship this and it is live, it will be difficult to change.

If we returned a versioned schema, we can handle that at the other end? 
Perhaps it is ok to leave now and we can do duck type checks at the other side when/if we do change this to return schema?

Steve-Mcl:
> > also, is get-node-type (singular) not a bit restrictive considering how the agent is put together?
> > Perhaps this should support an array of types like get-node-types ['function','ui-text'] (which can then be called with one element to support singular requests)
> 
> IMO having it singular or array doesn't make that much of a difference but please let me know how you guys would like to proceed. To have it standardised

I worry about lock in and never changing this. If we support plural up-front, improvements in the agent can be realised without re-deploying assistant to millions of customers.

andypalmi:
> also, is get-node-type (singular) not a bit restrictive considering how the agent is put together?
> Perhaps this should support an array of types like get-node-types ['function','ui-text'] (which can then be called with one element to support singular requests)

Done — renamed to automation/get-node-types and the types param now accepts an array. Returns an object keyed by type name so results are O(1) to look up. Types not installed get `{ installed: false }` rather than a top-level error, so one call can batch-check multiple types and handle partial misses cleanly.

@Steve-Mcl

resources/expertAutomations.js
```diff
@@ -1167,6 +1200,28 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+            for (const type of params.types) {
+                const def = this.RED.nodes.getType(type)
+                if (!def) {
+                    result.data[type] = { installed: false }
+                    continue
+                }
+                const rawDefaults = def.defaults || {}
+                result.data[type] = {
+                    defaults: JSON.parse(JSON.stringify(rawDefaults, (key, value) =>
+                        typeof value === 'function' ? value.toString() : value
```
Steve-Mcl: look at how node-red serialises Buffers/maps/sets etc for transmitting debug messages from BE to FE.

i am certain it converts them to an object something like:
```
{
   type: 'function',
   data: function.toString()
}
```

resources/expertAutomations.js
```diff
@@ -1167,6 +1200,28 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+                    )),
+                    label: typeof def.label === 'function' ? def.label.toString() : (def.label || type),
```
Steve-Mcl: same as above (regarding serialisation)

resources/expertAutomations.js
```diff
@@ -1167,6 +1200,28 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+                    category: def.category || null,
+                    color: typeof def.color === 'function' ? def.color.call({}) : (def.color || null),
```
Steve-Mcl: same as above (regarding serialisation)

Steve-Mcl (approved):

PR #297, opened 2026-04-29
fix(actions): allow config nodes in add-nodes without z property

### LINKED ISSUE #296: fix(actions): add-nodes rejects config nodes that have no z property
## Problem

The `addNodes` action handler unconditionally requires a `z` (tab ID) property on every node:

```js
if (!rawNode.z) throw new Error('Node is missing required property: z')
```

Config nodes (`category: 'config'`) like `ui-base`, `ui-theme`, `ui-page`, `ui-group` are global in Node-RED and have no `z` assignment. This hard guard rejects all config node creation via `add-nodes`, forcing the use of `import-flow` as a workaround.

The downstream `RED.view.importNodes()` call handles config nodes without `z` natively.

## Fix

- Check the node definition's `category` before enforcing `z` — skip the guard for config nodes
- Filter `undefined` values from the workspace validation loop to avoid `hasWorkspace(undefined)` throwing

andypalmi:
## Summary

- Config nodes (category `config`) like `ui-base`, `ui-theme`, `ui-page`, `ui-group` are global in Node-RED and have no `z` (tab) assignment
- The `addNodes` guard unconditionally required `z`, rejecting all config node creation
- Check the node definition's category via `RED.nodes.getType()` before enforcing `z`
- Filter `undefined` from the workspace validation loop so config nodes don't trigger `hasWorkspace(undefined)`

Closes #296

## Test plan

- Existing test updated: "should throw if non-config node is missing required property z" verifies non-config nodes still require `z`
- New test: "should accept config nodes without z property" verifies config nodes pass through to `importNodes` without `z`
- All 303 tests pass

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -244,10 +244,10 @@ export class ExpertAutomations extends ExpertActionsInterface {
                                 type: { type: 'string', description: 'Node type identifier' },
                                 x: { type: 'number', description: 'Canvas x position' },
                                 y: { type: 'number', description: 'Canvas y position' },
-                                z: { type: 'string', description: 'Tab (workspace) ID' }
+                                z: { type: 'string', description: 'Tab (workspace) ID — required for non-config nodes, omit for config nodes' }
                             },
                             additionalProperties: true,
-                            required: ['id', 'type', 'z']
+                            required: ['id', 'type']
                         },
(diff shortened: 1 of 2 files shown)


Steve-Mcl (approved):
LGTM

Steve-Mcl (approved):

PR #323, opened 2026-05-11
feat(automation): add automation/manage-groups action

andypalmi:
## Summary

Adds `automation/manage-groups` action with an operations array supporting four ops:

- `create`: groups a list of node IDs together (all must be on the same tab); uses `core:group-selection` internally. Accepts optional `name`, `style`, `env`, and explicit `id`.
- `update`: changes group metadata — `name`, `style`, `env`. Does not touch member nodes.
- `manage-members`: adds or removes nodes from an existing group using `mode: "add" | "remove"`.
- `delete`: disbands the group container via `core:ungroup-selection`; member nodes are kept in place.

Groups cannot be created via `add_nodes` — that action returns a clear error directing to `manage_groups`.

Defensive `_verifySelection` check applied to all four operations: after `RED.view.select()`, the actual canvas selection is verified against the expected nodes and throws if they don't match, guarding against silent selection failures in unexpected editor states.

## Test plan

- create: valid group, optional name/style/env, explicit id, id already in use, empty nodeIds, mixed-tab nodes, node not found, workspace locked, selection mismatch
- update: rename, style change, env replace, group not found, missing id, workspace locked
- manage-members: add nodes, remove nodes, invalid mode, missing id, node not in group, node on wrong tab
- delete: valid delete, group not found, missing id, workspace locked
- unknown op and empty operations array
- multiple operations in a single call

Closes #317

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -25,6 +25,7 @@ const GET_NODE_TYPES = 'automation/get-node-types'
 const GET_PALETTE = 'automation/get-palette'
 const LIST_CONFIG_NODES = 'automation/list-config-nodes'
 const OPEN_PALETTE_MANAGER = 'automation/open-palette-manager'
+const MANAGE_GROUPS = 'automation/manage-groups'
 
 /**
  * @typedef {SELECT_NODES
@@ -50,7 +51,8 @@ const OPEN_PALETTE_MANAGER = 'automation/open-palette-manager'
  *   |GET_NODE_TYPES
  *   |GET_PALETTE
  *   |LIST_CONFIG_NODES
- *   |OPEN_PALETTE_MANAGER} ExpertAutomationsActionsEnum
+ *   |OPEN_PALETTE_MANAGER
+ *   |MANAGE_GROUPS} ExpertAutomationsActionsEnum
  */
 
 export class ExpertAutomations extends ExpertActionsInterface {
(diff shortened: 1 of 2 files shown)


resources/expertAutomations.js
```diff
@@ -377,6 +379,59 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+                                    type: 'object',
+                                    description: 'Visual style: { fill, stroke, color, fontSize, bold, italic, label }',
+                                    properties: {
+                                        fill: { type: 'string' },
+                                        stroke: { type: 'string' },
+                                        color: { type: 'string' },
+                                        fontSize: { type: 'number' },
+                                        bold: { type: 'boolean' },
+                                        italic: { type: 'boolean' },
+                                        label: { type: 'string' }
```
Steve-Mcl: not matching actual group node spec

resources/expertAutomations.js
```diff
@@ -1411,6 +1506,151 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+                        group.nodes.push(node)
+                        node.g = group.id
+                    }
+                }
+            }
+        }
+
+        if (style) {
+            if (!group.style) group.style = {}
+            changes.style = { ...group.style }
```
Steve-Mcl, line 1672: should be explicit

resources/expertAutomations.js
```diff
@@ -1411,6 +1506,151 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+        const group = this.RED.nodes.group(id)
+        if (!group) throw new Error(`Group ${id} not found`)
+        if (group.z && this.RED.workspaces.isLocked(group.z)) {
+            throw new Error(`Workspace tab ${group.z} is locked`)
+        }
+        this.RED.history.push({ t: 'delete', nodes: [group], links: [], dirty: this.RED.nodes.dirty() })
+        this.RED.view.select({ nodes: [group] })
+        this.RED.actions.invoke('core:ungroup-selection')
+        this.RED.nodes.dirty(true)
+        this.RED.view.redraw()
```
Steve-Mcl, line 1704: core action already takes care of undo stack

resources/expertAutomations.js
```diff
@@ -1411,6 +1513,165 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+    /**
+     * Delete a group (ungroup its nodes).
+     * @param {string} id - group ID
+     */
+    deleteGroup (id) {
+        const group = this.RED.nodes.group(id)
+        if (!group) throw new Error(`Group ${id} not found`)
+        this._assertWorkspaceNotLocked(group.z)
+        this.RED.view.select({ nodes: [group] })
+        this.RED.actions.invoke('core:ungroup-selection')
```
Steve-Mcl, line 1702: after selecting & before performing operations, it would be sensible? (defensive) to call "get selection" and verify that ONLY the _items of interest_ were actually selected (the selection operation might have been prevented by some odd editor state!)

Steve-Mcl (approved):
on the surface, this appears to cover all angles - lets move forward with this - we can iterate on the internal logic as needed

PR #326, opened 2026-05-11
feat(automation): hard-reject group type in add-nodes, update-node, and remove-nodes

andypalmi:
## Summary

- `add-nodes`: rejects any node with `type: group` — groups must be created via `automation/manage-groups`
- `update-node`: rejects calls where the target ID is a group — use `automation/manage-groups` to modify groups
- `remove-nodes`: rejects any ID that resolves to a group — use `automation/manage-groups` to delete groups

In all three cases the action returns a clear error message directing to the correct action rather than silently failing or producing an invalid canvas state.

## Test plan

- `add-nodes`: passing `type: group` throws and directs to `manage-groups`
- `update-node`: passing a group ID returns `success: false` with a `manage-groups` error message
- `remove-nodes`: passing a group ID throws and directs to `manage-groups`

Closes #321

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -1260,6 +1260,11 @@ export class ExpertAutomations extends ExpertActionsInterface {
             break
 
         case UPDATE_NODE: {
+            if (this.RED.nodes.group(params.id)) {
+                result.error = 'Groups cannot be modified via update-node. Use automation/manage-groups instead.'
+                result.success = false
+                break
+            }
             await this.updateNode(params.id, params.properties, params.patches)
             const updatedNode = this.RED.nodes.node(params.id)
             result.data = this._summarizeNode(updatedNode)
@@ -1331,6 +1336,12 @@ export class ExpertAutomations extends ExpertActionsInterface {
             break
 
         case ADD_NODES: {
+            const groupNodes = (params.nodes || []).filter(n => n.type === 'group')
+            if (groupNodes.length > 0) {
+                result.error = 'Groups cannot be created via add-nodes. Use automation/manage-groups instead.'
+                result.success = false
+                break
+            }
             this.addNodes(params.nodes, { generateIds: params.generateIds ?? false })
             const addedNodes = params.nodes.map(n => this.RED.nodes.node(n.id)).filter(Boolean)
             if (this.RED.editor?.validateNode) {
@@ -1342,10 +1353,17 @@ export class ExpertAutomations extends ExpertActionsInterface {
         }
             break
 
-        case REMOVE_NODES:
+        case REMOVE_NODES: {
+            const groupIds = (params.ids || []).filter(id => !!this.RED.nodes.group(id))
+            if (groupIds.length > 0) {
+                result.error = `Groups cannot be removed via remove-nodes. Use automation/manage-groups with op: delete instead. Group IDs: ${groupIds.join(', ')}`
+                result.success = false
+                break
+            }
             this.removeNodes(params.ids)
             result.data = { removed: params.ids }
             result.success = true
+        }
             break
 
         case SET_WIRES:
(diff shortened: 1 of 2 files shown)


resources/expertAutomations.js
```diff
@@ -1260,6 +1260,11 @@ export class ExpertAutomations extends ExpertActionsInterface {
             break
 
         case UPDATE_NODE: {
+            if (this.RED.nodes.group(params.id)) {
+                result.error = 'Groups cannot be modified via update-node. Use automation/manage-groups instead.'
```
```suggestion
                result.error = `Groups cannot be modified via ${UPDATE_NODE}`
```

I think the "do other automation" suggestion should be handled in MCP tools / AI Agent. This is where developing consistent error codes would be useful 

Happy to discuss either way.

resources/expertAutomations.js
```diff
@@ -1331,6 +1336,12 @@ export class ExpertAutomations extends ExpertActionsInterface {
             break
 
         case ADD_NODES: {
+            const groupNodes = (params.nodes || []).filter(n => n.type === 'group')
+            if (groupNodes.length > 0) {
+                result.error = 'Groups cannot be created via add-nodes. Use automation/manage-groups instead.'
```
Steve-Mcl: again, these suggestions are not really of use to agent due to name alignment - equally, we dont want to advertise the real backend info here - so this warrants structured error codes that the MCP tools and/or agent understands (or some other solution)

Lets discuss options and merits tomorrow.

resources/expertAutomations.js
```diff
@@ -1342,10 +1353,17 @@ export class ExpertAutomations extends ExpertActionsInterface {
         }
             break
 
-        case REMOVE_NODES:
+        case REMOVE_NODES: {
+            const groupIds = (params.ids || []).filter(id => !!this.RED.nodes.group(id))
+            if (groupIds.length > 0) {
+                result.error = `Groups cannot be removed via remove-nodes. Use automation/manage-groups with op: delete instead. Group IDs: ${groupIds.join(', ')}`
```
Steve-Mcl: same again

Steve-Mcl (commented):
We should discuss the merits of offering alternative "do the other command" vs structured errors

resources/expertAutomations.js
```diff
@@ -1335,6 +1352,20 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+                result.success = false
+                break
+            }
+            const gNode = (params.nodes || []).find(n => n.g !== undefined)
+            if (gNode) {
+                result.error = `Node ${gNode.id}: "g" cannot be set directly — group membership must be managed via a dedicated action`
+                result.errorCode = ERROR_CODES.FORBIDDEN_PROPERTY
+                result.success = false
+                break
+            }
```
Steve-Mcl, line 1374: I wonder if ADD_NODES should allow a `g` prop - I mean its OK to add a "new node" to an "existing group" (so long the group exists) - WDYT?

resources/expertAutomations.js
```diff
@@ -1346,10 +1377,18 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
 
-        case REMOVE_NODES:
+        case REMOVE_NODES: {
+            const groupIds = (params.ids || []).filter(id => !!this.RED.nodes.group(id))
+            if (groupIds.length > 0) {
+                result.error = `IDs [${groupIds.join(', ')}] are group nodes — group nodes cannot be removed via this action`
+                result.errorCode = ERROR_CODES.GROUP_OPERATION_REQUIRED
+                result.success = false
+                break
+            }
```
Steve-Mcl, line 1393: I think remove node being called for a group should actually remove the group and all nodes - thats what node-red editor does (kinda - it actually selects all nodes with matching `g` then deletes selection)

I am torn!

Steve-Mcl (approved):
I think we could improve, but I am certain this can be retro'd later so wont block - but also, I wont merge until we have a quick chat (in case we really do want to tweak this)

Steve-Mcl (changes requested):
After discussion, we will address #331 in this PR

Steve-Mcl (approved):

PR #327, opened 2026-05-11
feat(automation): slim get-workspace-nodes response by default

### LINKED ISSUE #316: Slim automation/get-workspace-nodes response by default
## Problem

`automation/get-workspace-nodes` returns the full node set from `RED.nodes.createCompleteNodeSet()`, including all node properties (wires, credentials, full config, etc.). For routine canvas inspection this produces an excessive payload — often thousands of lines — when only a handful of fields are needed.

## Proposed change

Return slim node summaries by default (id, type, name, x, y, z, valid) using the existing `_summarizeNode` and `_summarizeGroup` helpers, extended with a new `_summarizeSubflow` helper. A `full: true` parameter opts into the complete payload for cases where every property is genuinely needed.

## Acceptance criteria

- Default response returns slim summaries for tabs, nodes, groups, and subflow definitions
- Passing `full: true` returns the unmodified complete node set
- `_summarizeSubflow` exposes input/output port counts rather than the raw `in`/`out` port arrays
- Existing tests updated to reflect slim-by-default behaviour

andypalmi:
## Summary

`automation/get-workspace-nodes` now returns a condensed summary of each flow item by default instead of the full node object:

- Tabs: `id`, `type`, `label`, `disabled`
- Subflows: `id`, `type`, `name`, `info`, `inputs`, `outputs`
- Groups: `id`, `type`, `name`, `z`, `nodeCount`, `style`, `env`
- All other nodes: `id`, `type`, `name`, `label`, `x`, `y`, `z`, `valid`

Pass `full: true` to get the complete node objects as before.

## Test plan

- Default call returns slim summaries for tabs, subflows, groups, and regular nodes
- `full: true` returns raw node data unchanged
- Empty flow returns empty array

Closes #316

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -1260,6 +1260,11 @@ export class ExpertAutomations extends ExpertActionsInterface {
             break
 
         case UPDATE_NODE: {
+            if (this.RED.nodes.group(params.id)) {
+                result.error = 'Groups cannot be modified via update-node. Use automation/manage-groups instead.'
+                result.success = false
+                break
+            }
             await this.updateNode(params.id, params.properties, params.patches)
             const updatedNode = this.RED.nodes.node(params.id)
             result.data = this._summarizeNode(updatedNode)
@@ -1331,6 +1336,12 @@ export class ExpertAutomations extends ExpertActionsInterface {
             break
 
         case ADD_NODES: {
+            const groupNodes = (params.nodes || []).filter(n => n.type === 'group')
+            if (groupNodes.length > 0) {
+                result.error = 'Groups cannot be created via add-nodes. Use automation/manage-groups instead.'
+                result.success = false
+                break
+            }
             this.addNodes(params.nodes, { generateIds: params.generateIds ?? false })
             const addedNodes = params.nodes.map(n => this.RED.nodes.node(n.id)).filter(Boolean)
             if (this.RED.editor?.validateNode) {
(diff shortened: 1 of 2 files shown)


resources/expertAutomations.js
Steve-Mcl, line 194: should there not be an optional `full` property in the spec?

resources/expertAutomations.js
```diff
@@ -1279,15 +1279,18 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
                     // filter by type if specified (e.g. "tab", "subflow", or any node type)
                     result.flows = result.flows.filter(f => f.type === params.type)
                 }
-                if (params.tabId) {
+                if (params && params.tabId) {
                     // filter by parent tab ID if specified (for nodes/config nodes)
                     result.flows = result.flows.filter(f => f.z === params.tabId)
                 }
             }
+            if (!params || !params.full) {
```
Steve-Mcl, line 1318: full doesnt seem to be decared.

Steve-Mcl (commented):
please check the spec for GET_FLOW - in particular the new full option

resources/expertAutomations.js
andypalmi: Added it in https://github.com/FlowFuse/nr-assistant/pull/327/changes/413655673f937a9d5ed69fb5c3772d7222e72417

resources/expertAutomations.js
andypalmi: Added it in https://github.com/FlowFuse/nr-assistant/pull/327/changes/413655673f937a9d5ed69fb5c3772d7222e72417

Steve-Mcl (approved):

PR #328, opened 2026-05-11
feat(automation): hard-reject wires property in update-node

### LINKED ISSUE #320: Hard validation: reject wire properties in update-node
## Problem

`automation/update-node` accepts an arbitrary properties map, which means callers can attempt to set `wires` or `links` directly on a node. This bypasses the dedicated wire management actions and can produce inconsistent canvas state, since wires in Node-RED are bidirectional relationships that must be maintained through the proper APIs.

## Proposed change

Reject any update that includes `wires` or `links` as a property key. Return a clear error message directing the caller to use `automation/set-wires` or `automation/set-links` instead. This mirrors the validation already present in `automation/add-nodes` via `FORBIDDEN_NODE_PROPERTIES`.

## Acceptance criteria

- Attempting to set `wires` or `links` via `update-node` returns an error
- The error message names the offending property and names the correct alternative action
- Unit test covers the rejection path

andypalmi:
Closes #320

## Summary

- `automation/update-node`: throws if `properties` contains a `wires` key, or a `links` key when the node is of type `link in`, `link out`, or `link call` — connections must go through dedicated actions
- `automation/add-nodes`: same rejection — throws if any node definition has a `wires` key, or a `links` key on a link node type

## Test plan

- `update-node`: passing `properties: { wires: [...] }` throws with a message directing to the dedicated action
- `update-node`: passing `properties: { links: [...] }` on a `link out` node throws; on a non-link node it is allowed
- `add-nodes`: passing a node with a `wires` property throws with the same direction
- `add-nodes`: passing a `link out` node with a `links` property throws; a non-link node with `links` is allowed

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -32,6 +32,8 @@ const ERROR_CODES = Object.freeze({
     FORBIDDEN_PROPERTY: 'FORBIDDEN_PROPERTY'
 })
 
+const LINK_NODE_TYPES = ['link in', 'link out', 'link call']
+
 /**
  * @typedef {SELECT_NODES
  *   |GET_NODES
@@ -1287,7 +1289,7 @@ export class ExpertAutomations extends ExpertActionsInterface {
                 result.success = false
                 break
             }
-            if (params.properties && 'links' in params.properties) {
+            if (params.properties && 'links' in params.properties && LINK_NODE_TYPES.includes(this.RED.nodes.node(params.id)?.type)) {
                 result.error = `Node ${params.id}: "links" cannot be set directly — link connections must be managed via a dedicated action`
                 result.errorCode = ERROR_CODES.FORBIDDEN_PROPERTY
                 result.success = false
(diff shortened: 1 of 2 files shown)


Steve-Mcl (commented):
Can we duicuss the whole "use automation/xxx" appraoch before we go down this path. In short, I suspect since names on assistant do not align perfectly with the backend, this might confuse AI. We might want to consider structured error codes that can be used in backend?

Steve-Mcl (commented):
Lets also check that update node uses the link management for link nodes

resources/expertAutomations.js
```diff
@@ -1281,6 +1281,18 @@ export class ExpertAutomations extends ExpertActionsInterface {
                 result.success = false
                 break
             }
+            if (params.properties && 'wires' in params.properties) {
+                result.error = `Node ${params.id}: "wires" cannot be set directly — wire connections must be managed via a dedicated action`
+                result.errorCode = ERROR_CODES.FORBIDDEN_PROPERTY
+                result.success = false
+                break
+            }
+            if (params.properties && 'links' in params.properties) {
```
Steve-Mcl: We need to be explicit that this is a core link node (other contrib nodes may legit have a links prop!)

Steve-Mcl (approved):

PR #329, opened 2026-05-11
feat(automation): migrate update-node to update-nodes (array params)

andypalmi:
Closes #325

## Summary

- Renames `automation/update-node` to `automation/update-nodes`
- Replaces the separate `properties`/`patches` params with a single `updates` array — each item targets a property by name and supports two modes:
  - Full replacement: `{ property: "name", op: "replace", content: "My Node" }` — omit `start`/`end`
  - Line-based edit: `{ property: "func", op: "replace", start: 1, end: 3, content: "new code" }` — provide `start` (and `end` for replace/delete)
- The action now does the full/line split internally, matching the MCP tool input schema exactly — `start` is optional throughout
- Accepts multiple node entries in one call; result changes from `{ data: {...}, validation: {...} }` to `{ data: [{ ...summary, validation }] }` — one entry per node
- Adds wires, links (link-type nodes only), and g rejection checks, consistent with `add-nodes`
- MCP tool updated to pass `updates` directly without pre-splitting

## Test plan

- All property update tests migrated to `updates` array format
- All line-based patch tests migrated to `updates` array format with `start`
- Full replacement and line-based edits combined in one call
- Group rejection, wires rejection, links rejection (link node types only), and g rejection tests updated

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -6,7 +6,7 @@ const GET_NODES = 'automation/get-nodes'
 const EDIT_NODE = 'automation/open-node-edit'
 const SEARCH = 'automation/search'
 const ADD_FLOW_TAB = 'automation/add-flow-tab'
-const UPDATE_NODE = 'automation/update-node'
+const UPDATE_NODES = 'automation/update-nodes'
 const SHOW_WORKSPACE = 'automation/show-workspace'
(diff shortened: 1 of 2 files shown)


Steve-Mcl (approved):
tested and working!

PR #341, opened 2026-05-12
fix(automation): track history for moveNodeToTab so undo works correctly

andypalmi:
Closes #337

## Summary

`_moveNodeToTab` called `RED.nodes.moveNodeToTab` but never pushed a history event for the operation. The wire-splitting steps (junction insertion, link-node creation) each push their own history events via the core actions they invoke, but the actual node relocation had no entry in the undo stack. Pressing Ctrl+Z would undo the wire splits while leaving the nodes stranded on the target tab.

The fix pushes a `multi` history event with a stateful toggle callback after all `moveNodeToTab` calls complete. The callback moves each node (and its adjacent link nodes) back to their original tab on undo, then swaps the recorded source/target so the same callback correctly re-applies the move on redo.

Closes https://github.com/FlowFuse/nr-assistant/issues/337

## Test plan

- Existing tab-change tests updated to include `mockRED.history` in setup
- New test: verifies a `multi` event with the correct callback is pushed when moving a node with no wires, and that calling the callback alternates between moving the node to the target tab and back
- New test: verifies the callback includes all link nodes created during wire splitting, and moves them back to the source tab on undo


cstns (approved):
Looks good to me. We can quickly iterate if we missed anything

PR #347, opened 2026-05-20
feat(automation): add align-selection action

andypalmi:
## Summary
- Add `automation/align-selection` action that reveals the workspace, selects nodes by ID, and invokes `core:align-selection-to-*` actions
- Supports grid, left, right, top, bottom, middle, and center alignment directions
- Validates minimum 2 non-config nodes for non-grid directions
- Config nodes are automatically excluded from the selection since they lack canvas positions and break alignment
- Returns structured result with `alignedNodes` and `excludedConfigNodes` so the caller knows which nodes were skipped
- Workspace existence and lock checks are performed before alignment

Closes #334


cstns (approved):
changes look good to me, please address the linting errors before merging

PR #349, opened 2026-05-20
feat(automation): support include with multiple node IDs

andypalmi:
## Summary

- The getNodes and selectNodes methods now support the include parameter (upstream, downstream, connected) when called with an array of IDs, iterating each node's connections and merging results with deduplication
- Extracted _getConnectedNodes helper to share traversal logic between single-node and multi-node code paths
- getNodes now always returns an array and throws if any node ID is not found, replacing null-return behavior
- Defensive guards retained in invokeAction for both null and empty results

## Test plan

- Verify include with multiple IDs returns connected nodes from all starting points, deduplicated
- Verify overlapping connections across IDs are properly deduplicated
- Verify not-found node IDs throw descriptive errors
- Verify empty results still throw descriptive errors in the action dispatcher

Closes #348


resources/expertAutomations.js
```diff
@@ -483,49 +483,67 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+        for (const node of nodes) {
+            const connected = this._getConnectedNodes(node, include)
+            for (const n of connected) {
+                if (!seen.has(n.id)) {
+                    seen.add(n.id)
+                    result.push(n)
+                }
+            }
+        }
+        return result
```
Steve-Mcl: This will return an odd shaped array.
Also, IIRC, get upstream includes the source node so we might get duplicates in the result.

I think a better approach is to 
* loop `ids` - for each `id`
    - if `id` is already collected, `continue` (skip performing the include op altogether for this node)
    - get the `include` (if any)
    - loop the result of `_getConnectedNodes` & only add if not existing

Steve-Mcl (commented):
As discussed, I still really dislike this. I think the existing arrangement of splitting the action at the MCP server into 2 distinct tools of `select_node` (with include) and `select_nodes` (for selecting a list of ids) is the better approach - but I wont block. 

1 last consideration - have you checked the affect of throwing errors has (instead of the original behaviour of returning null/empty array) on the existing expert action links (e.g. `[function 1](nr-assistant://automation/select-nodes?id=709483aa8eeae4e5e)` / `[function 2](nr-assistant://automation/open-node-edit?id=709483aa8eeae4e5e)`

PR #351, opened 2026-05-20
feat(automation): add leveled connection traversal

andypalmi:
## Summary

- Add `levels` parameter to get-nodes and select-nodes actions for BFS traversal
- When `levels` is specified with `include`, nodes are returned grouped by connection level (0 = start, 1 = direct neighbors, etc.)
- `levels=0` traverses all levels; omitting `levels` preserves flat-array behavior
- Add `includeConfigNodes` parameter (default false) to control whether config nodes appear in traversal results
- Filter config nodes injected by `createExportableNodeSet` during formatting, not just during BFS traversal

## Test plan

- Verify downstream traversal returns nodes grouped by correct levels
- Verify maxLevels limits traversal depth (levels=1 returns only direct neighbors)
- Verify upstream traversal uses input links correctly
- Verify connected traversal walks both directions
- Verify deduplication across multiple start nodes
- Verify existing flat-array behavior when levels is omitted
- Verify config nodes are excluded from results when includeConfigNodes is false (default)
- Verify config nodes injected by createExportableNodeSet are filtered post-formatting

Closes #350

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -92,6 +92,11 @@ export class ExpertAutomations extends ExpertActionsInterface {
                         type: 'integer',
                         minimum: 0,
                         description: 'Max connection levels to traverse when include is specified. 1 = direct neighbors only, 2 = two hops, etc. 0 = all levels (no limit). Results are grouped by level. Only applicable when include is provided.'
+                    },
+                    includeConfigNodes: {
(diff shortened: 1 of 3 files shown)


andypalmi:
Closing — decided to rely on Node-RED core actions instead of maintaining custom traversal code

PR #363, opened 2026-05-26
feat(automations): add export-flow action

andypalmi:
Adds a new `automation/export-flow` action that opens the Node-RED export dialog pre-scoped to the requested range.

The action accepts a `scope` parameter:
- `selection` — exports the currently selected nodes
- `current-tab` — exports all nodes on the active flow tab
- `all-flows` — exports the entire workspace

Internally it invokes `core:show-export-dialog` and then clicks the corresponding range button in the dialog DOM so the correct scope is pre-selected when the dialog opens.

Closes [#359](https://github.com/FlowFuse/nr-assistant/issues/359)

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -497,7 +497,7 @@ export class ExpertAutomations extends ExpertActionsInterface {
                     },
                     tabId: {
                         type: 'string',
-                        description: 'ID of the tab to switch to before exporting. Only valid when scope is "current-tab". The tab must exist or an error is thrown.'
+                        description: '(optional) The ID of the tab to switch to before exporting. Only valid when scope is "current-tab". If the tab does not exist, an error will be thrown.'
                     }
                 },
                 required: ['scope']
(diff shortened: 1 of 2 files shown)


resources/expertAutomations.js
```diff
@@ -484,6 +486,23 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+                type: 'object',
+                properties: {
+                    scope: {
+                        type: 'string',
+                        enum: ['selection', 'current-tab', 'all-flows'],
+                        description: '"selection" exports the currently selected nodes; "current-tab" exports all nodes on the active flow tab; "all-flows" exports the entire workspace'
+                    },
+                    tabId: {
+                        type: 'string',
+                        description: 'ID of the tab to switch to before exporting. Only valid when scope is "current-tab". The tab must exist or an error is thrown.'
```
```suggestion
                        description: '(optional) The ID of the tab to switch to before exporting. Only valid when scope is "current-tab". If the tab soes not exist, an error will be thrown.'
```

resources/expertAutomations.js
```diff
@@ -484,6 +486,23 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
             }
         },
+        [EXPORT_FLOW]: {
+            params: {
+                type: 'object',
+                properties: {
+                    scope: {
+                        type: 'string',
+                        enum: ['selection', 'current-tab', 'all-flows'],
+                        description: '"selection" exports the currently selected nodes; "current-tab" exports all nodes on the active flow tab; "all-flows" exports the entire workspace'
```
```suggestion
                        description: 'Scope specifies what should be exported. Options are "selection" to export currently selected nodes, "current-tab" to export all nodes on the active flow tab, "all-flows" to export all flows on all workspace tabs including config nodes and subflows'
```

Steve-Mcl (approved):

PR #371, opened 2026-06-24
feat: add create-subroutine automation action

andypalmi:
## What

Adds a new `automation/create-subroutine` action that wraps one or more existing nodes into a callable **link-node subroutine**, replacing them inline with a single `link call`, and builds it as a self-contained, reusable unit:

```
caller:  <upstream> -> [link call] -> [switch on msg.error] --Success--> <downstream>
                                            \--Error--> [debug]
body:    [link in] -> <entry> ... <exit> -> [link out, mode: return]
         [catch (group scope)] -> [link out]
```

The link call targets the link in, so at runtime the message is routed into the body and returned to the caller by the return-mode link out. The body (link in, the selected nodes, link out and the catch) is wrapped in a named group, a group-scoped catch returns body errors through the same link out, and a switch on `msg.error` splits the returned message into a labelled **Error** branch and **Success** branch on the caller side. The caller (link call, switch, debug) stays on the original row where the selection was; the body group is placed just below it.

## Why

There is no Node-RED core action for the link-call (call/return) subroutine pattern: `core:split-wire-with-link-nodes` only produces a one-way bridge, and `core:create-subflow` produces a subflow. Unlike a subflow, which duplicates nodes for every instance, a link-call subroutine is a single shared body that any caller reuses. Building the pattern by hand takes several fragile steps (create link in/out/call, set the link out to return mode, rewire the caller, add the catch and switch, group the body, set the virtual link). This action turns it into a single, validated call.

## How

- The selection must form a **single-entry, single-exit subgraph**: a single node, or connected nodes with one way in and one way out (this includes a branch that reconverges, not only a straight chain). Multi-entry / multi-exit selections are rejected, since a `link call` has a single input and a single output. External wires that enter a non-entry node or leave a non-exit node are also rejected, so the body cannot be entered or left except through its single entry and exit.
- A **group ID is accepted and expanded** to its member nodes (recursively for nested groups), so a whole group can be wrapped in one call; the expanded nodes are then validated like any other selection. Nodes that are already group members (whether a group ID was passed, or node IDs whose nodes live in a group) are **detached from their group before being re-grouped** into the subroutine, since Node-RED's group-selection only groups nodes that share the same parent group. Any original group the detach empties is removed so no empty group is left behind, while a group that still holds other members is left intact.
- Any ordinary canvas node can be wrapped, including subflow instances and junctions. Config and link nodes are rejected (config nodes have no canvas position; link nodes are the subroutine's own plumbing).
- Internal wires between selected nodes are kept intact. External inbound is redirected to the link call; the link call's return feeds a switch on `msg.error`. The **Success** branch continues to whatever was downstream, and the **Error** branch is wired to a `debug` node by default so failures are visible (the caller can rewire it to a notification, a log, or further handling).
- A group-scoped `catch` is wired to the link out, so an error raised anywhere in the body returns through the same path and surfaces on the switch's Error branch.
- The body is wrapped in a named group (named after the subroutine) so it reads as one unit on the canvas, and is dropped below the caller row so the group does not overlap the call.
- A `timeout` option sets the link call timeout in seconds, defaulting to 5.
- The action only **creates** a subroutine on the current tab. Relocating the result to a dedicated "Subroutines" tab, renaming it or deleting it is left to the existing node and group actions (move via a node `z` update, rename/regroup via the group action, delete via remove) rather than being bolted onto this action.
- Rather than poking `RED` primitives directly, it orchestrates the module's own validated actions (`addNodes` / `setWires` / `setLinks` / `createGroup`), so node-type, port, same-tab, link-type and duplicate-wire validation all apply.

## Tests

Adds unit coverage for the single-node case (including the group, catch, error switch, the `Error`/`Success` labels and the default Error-branch debug wiring), the supplied-timeout option, linear chains (internal wire left intact), group expansion, the empty-group rejection, and the multi-exit / external-wire-into-a-middle-node / link-node / config-node / empty-input rejections. Full suite: 437 passing.

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -345,10 +345,9 @@ export class ExpertAutomations extends ExpertActionsInterface {
                         type: 'array',
                         items: { type: 'string' },
                         minItems: 1,
-                        description: 'IDs of the existing node(s) to move into the subroutine body. Must form a single-entry, single-exit subgraph: a single node, or connected nodes with one way in and one way out (a branch that reconverges is fine, not only a straight chain).'
+                        description: 'IDs of the existing node(s) to move into the subroutine body. A group ID is accepted and expanded to its member nodes (nested groups recurse). The resulting nodes must form a single-entry, single-exit subgraph: a single node, or connected nodes with one way in and one way out (a branch that reconverges is fine, not only a straight chain).'
                     },
                     name: { type: 'string', description: 'Name for the subroutine, used to label the created link nodes and the group. Defaults to "Subroutine".' },
-                    targetTabId: { type: 'string', description: 'Optional tab ID to move the subroutine body (link in, the selected nodes, link out, catch) onto. The link call and error switch stay inline with the caller. Omit to keep the body on the same tab.' },
                     timeout: { type: 'number', description: 'Link call timeout in seconds. Defaults to 5.', default: 5 }
                 },
                 required: ['ids']
@@ -1511,14 +1510,17 @@ export class ExpertAutomations extends ExpertActionsInterface {
      * nodes must form a single-entry, single-exit subgraph (the common case is a
      * single node, or a linear chain). The result is:
      *   caller:  <upstream> -> [link call] -> [switch on msg.error] -> <downstream>
+     *                                              \-> [debug] (Error branch)
      *   body:    [link in] -> <entry> ... <exit> -> [link out, mode: return]
      *            [catch (group scope)] -> [link out]
      * The link call targets the link in, so at runtime the message is routed into
      * the body and returned to the caller by the return-mode link out. A group-scoped
      * catch routes body errors back out the same link out, and a switch on the
-     * caller side splits the returned message into an error branch and a normal
-     * branch. The body (link in, the selected nodes, link out and the catch) is
-     * wrapped in a named group.
+     * caller side splits the returned message into an "Error" branch and a "Success"
+     * branch; the Error branch is wired to a debug node so failures are visible by
+     * default. The link call, switch and debug stay on the original row where the
+     * selection was; the body (link in, the selected nodes, link out and the catch)
+     * is dropped below the caller and wrapped in a named group.
      *
      * Rather than hand-building node JSON, this orchestrates the module's own
      * validated actions (addNodes / setWires / setLinks / createGroup), so node-type,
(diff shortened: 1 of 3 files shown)


test/unit/resources/expertAutomations.test.js
```diff
@@ -828,6 +829,226 @@ describeMain('expertAutomations', () => {
...
+            })
+
+            it('uses the supplied timeout for the link call', () => {
+                setupSingleNode()
+                expertAutomations.createSubroutine({ ids: ['f1'], timeout: 30 })
+                const linkCall = expertAutomations.addNodes.firstCall.args[0].find(n => n.type === 'link call')
+                linkCall.timeout.should.equal('30')
+            })
+
+            it('rejects a config node in the selection (config nodes are global, with no x/y)', () => {
```
Steve-Mcl: config nodes in the selection are fine so long as you dont move them (the owner node is what is moved)
Essentially, ignore them

resources/expertAutomations.js
```diff
@@ -335,6 +337,22 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
         },
+        [CREATE_SUBROUTINE]: {
+            params: {
+                type: 'object',
+                properties: {
+                    ids: {
+                        type: 'array',
+                        items: { type: 'string' },
+                        minItems: 1,
+                        description: 'IDs of the existing node(s) to move into the subroutine body. A group ID is accepted and expanded to its member nodes (nested groups recurse). The resulting nodes must form a single-entry, single-exit subgraph: a single node, or connected nodes with one way in and one way out (a branch that reconverges is fine, not only a straight chain).'
```
Steve-Mcl: Can we include what this does (along side what is acceptable) e.g start with "Create a link-call subroutine."

resources/expertAutomations.js
```diff
@@ -335,6 +337,22 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+                    },
+                    name: { type: 'string', description: 'Name for the subroutine, used to label the created link nodes and the group. Defaults to "Subroutine".' },
```
```suggestion
                    name: { type: 'string', description: 'Name for the subroutine, used to label the created 'link in' node and the group name. Defaults to "Subroutine".' },
```

Steve-Mcl (commented):
couple of suggestions and a question

what happens if the original selection is a group already? Do we end up with a group in a group?

Steve-Mcl (approved):

PR #379, opened 2026-07-22
feat: add update-tab and remove-subflow automation actions

andypalmi:
## Summary
- Add an `automation/update-tab` action and `updateTab(id, properties)` method to rename a tab or change its enabled/disabled state, info, or env.
- Add an `automation/remove-subflow` action and `removeSubflow(id)` method that deletes a subflow definition together with its internal nodes and every placed instance, delegating to Node-RED's own subflow removal and pushing an undoable history event; blocked when an instance sits on a locked workspace.
- Fix `removeTab` to redraw the canvas after deleting a tab.
- Make `removeNodes` partial: existing nodes are removed and missing ids are reported, instead of failing the whole batch when one id is absent.
- Reject unknown property names on node updates, validated against the node type's declared defaults.
- Force a synchronous, cache-refreshed repaint after subflow removal, so a removed instance doesn't stay visible until an unrelated interaction (e.g. switching tabs) forces a redraw.
- Recognise junction nodes across `addNodes`, `isConfigNode`, `setWires`, `removeNodes`, `updateNode` and `_moveNodeToTab`. Junctions have no registered palette definition, so they live in `RED.nodes.junction` rather than the main node registry; `addNodes` was rejecting them outright with "Unknown node type: junction", and every other lookup would have failed the same way. Added a `_resolveNode` helper that falls back to `RED.nodes.junction`.

## Test plan
- [x] Full unit suite passing (`npm test`)
- [x] Live-validated against a running instance: add subflow, remove subflow, confirm both the definition and its placed instance are gone with no stale canvas state

Closes FlowFuse/engineering#176
Closes FlowFuse/engineering#174
Closes FlowFuse/engineering#177
Closes #378

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@flowfuse/nr-assistant",
-    "version": "0.18.0",
+    "version": "0.17.0",
     "description": "FlowFuse Node-RED Expert plugin",
     "main": "index.js",
     "scripts": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@flowfuse/nr-assistant",
-    "version": "0.17.0",
+    "version": "0.18.0",
     "description": "FlowFuse Node-RED Expert plugin",
     "main": "index.js",
     "scripts": {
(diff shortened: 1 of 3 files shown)


Steve-Mcl (approved):
tested locally

PR #384, opened 2026-07-30
fix: scope manage-groups operation errors to the failing operation

andypalmi:
### What
Wraps each `manage-groups` operation so a low-level error thrown by a Node-RED core group or wire action is re-thrown scoped to the operation and its target group.

Before, a failing operation propagated a raw core error (for example `Cannot read properties of undefined (reading 'has')`) with no indication of which operation or group caused it. Now the caller receives a message such as `manage-groups "move" operation failed for group "g1": ...`.

### Why
The per-operation helpers drive core editor actions programmatically, and those actions can throw when the editor state is not what core expects. Scoping the error makes the failure legible and diagnosable instead of surfacing an opaque core detail.

### Notes
This does not change the outcome of a successful operation, and it does not attempt to fix the underlying core throw. It only makes the failure legible. Fixes #383.

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -2343,49 +2343,58 @@ export class ExpertAutomations extends ExpertActionsInterface {
             for (const entry of operations) {
                 const op = entry.op
                 if (!op) throw new Error('op is required for each manage-groups operation')
-                switch (op) {
-                case 'create': {
-                    const group = this.createGroup(entry.nodeIds, entry.name, { style: entry.style, env: entry.env, id: entry.id })
-                    results.push({ op: 'create', ...this._summarizeGroup(group) })
-                    break
-                }
-                case 'update': {
-                    if (!entry.id) throw new Error('id is required for update')
-                    const group = this.updateGroup(entry.id, {
-                        name: entry.name,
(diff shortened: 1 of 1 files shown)


resources/expertAutomations.js
```diff
@@ -2343,49 +2343,58 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+                    default:
+                        throw new Error(`Unknown manage-groups op: ${op}`)
+                    }
+                } catch (err) {
+                    // The per-op helpers drive Node-RED core group/wire actions, which can throw
+                    // low-level errors (e.g. reading 'has' on a core selection Set) when the editor
+                    // state isn't what core expects. Re-throw scoped to the operation and its target
+                    // so the caller gets a legible message naming what failed, not a raw core detail.
+                    const target = entry.id ? ` for group "${entry.id}"` : (entry.name ? ` for "${entry.name}"` : '')
+                    throw new Error(`manage-groups "${op}" operation failed${target}: ${err.message}`)
```
```suggestion
                    throw new Error(`manage-groups "${op}" operation failed "${target}": ${err.message}`)
```

Steve-Mcl (approved):

PR #387, opened 2026-07-30
feat: add get-debug-messages automation action

### LINKED ISSUE #386: Add get-debug-messages automation action
## Summary

The Expert has no automation action for reading the Node-RED debug sidebar, so it cannot answer questions about recent debug/error output. The only existing mechanism (`debug-log-context-get-entries`) is wired as a raw postMessage command for the "add to chat context" UI feature, not a schema-validated `automation/*` action.

## Changes needed

- Add `automation/get-debug-messages` action returning the most recent debug sidebar entries, filterable by level (fatal/error/warn/info/debug/trace) and by source node id(s)
- Cap results at the sidebar's own 100-message retention (it doesn't keep more than that itself)
- Reuse the existing debug-message formatting/scraping built for "add to chat context" rather than duplicating it

andypalmi:
## What

Adds a new `automation/get-debug-messages` action that returns the most recent messages from the Node-RED debug sidebar, optionally filtered by level and by source node id(s).

## Why

The Expert has no way to read what's in the debug sidebar today. The only existing mechanism (`debug-log-context-get-entries`) is wired as a raw postMessage command for the "add to chat context" UI feature, not a schema-validated `automation/*` action.

## How

- Extracts the DOM-scraping logic already used by `handleDebugLogContextGetEntries` into a shared `collectDebugLogEntries({ visibleOnly })` method on `ExpertComms`, reused by both the existing chat-context command and the new action.
- Adds `getDebugMessages(params)` on `ExpertAutomations`: filters `collectDebugLogEntries` output by level (`fatal/error/warn/info/debug/trace`, all default true) and by `nodeIds` (matching either the message's source node or any ancestor, e.g. a subflow instance), then takes the last `count` entries (default 20, hard-capped at 100 since that's all the sidebar itself retains).
- Registers the action and its params schema in the `actions` map, and dispatches it in `invokeAction`.

## Tests

Adds unit coverage for count default/capping, level filtering, `nodeIds` filtering (direct source and ancestor match), and the dispatched action. Full suite: 472 passing.

Closes #386

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -34,6 +34,7 @@ const EXPORT_FLOW = 'automation/export-flow'
 const SET_DEPLOY_MODE = 'automation/set-deploy-mode'
 const SHOW_SIDEBAR_PANEL = 'automation/show-sidebar-panel'
 const TOGGLE_SIDEBAR = 'automation/toggle-sidebar'
+const GET_DEBUG_MESSAGES = 'automation/get-debug-messages'
 
 const ALIGNMENT_DIRECTIONS = ['grid', 'left', 'right', 'top', 'bottom', 'middle', 'center']
 const DISTRIBUTE_DIRECTIONS = ['horizontally', 'vertically']
@@ -615,6 +616,34 @@ export class ExpertAutomations extends ExpertActionsInterface {
                 },
                 required: ['side']
             }
+        },
+        [GET_DEBUG_MESSAGES]: {
+            params: {
+                type: 'object',
+                properties: {
+                    count: {
+                        type: 'number',
+                        default: 20,
+                        description: 'Maximum number of debug messages to return. Defaults to 20. The Node-RED debug sidebar itself only ever retains the last 100 messages, so this is capped at 100 regardless of the value provided.'
+                    },
+                    nodeIds: {
+                        type: 'array',
+                        items: { type: 'string' },
+                        description: 'Optional. Only include messages sourced from one of these node ids, or from a node nested within one of these ids (e.g. a node inside a subflow instance listed here).'
+                    },
+                    fatal: { type: 'boolean', default: true, description: 'Include "fatal" level messages. Defaults to true.' },
+                    error: { type: 'boolean', default: true, description: 'Include "error" level messages. Defaults to true.' },
+                    warn: { type: 'boolean', default: true, description: 'Include "warn" level messages. Defaults to true.' },
+                    info: { type: 'boolean', default: true, description: 'Include "info" level messages. Defaults to true.' },
+                    debug: { type: 'boolean', default: true, description: 'Include "debug" level messages. Defaults to true.' },
+                    trace: { type: 'boolean', default: true, description: 'Include "trace" level messages. Defaults to true.' },
+                    select: {
+                        type: 'boolean',
+                        default: false,
+                        description: 'If true, also select the source nodes of the returned messages on the canvas. Defaults to false.'
+                    }
+                }
+            }
         }
 
     })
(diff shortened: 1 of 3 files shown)


resources/expertAutomations.js
```diff
@@ -667,6 +696,59 @@ export class ExpertAutomations extends ExpertActionsInterface {
...
+            })
+        }
+
+        const maxCount = Math.min(Math.max(Number(count) || 20, 1), 100)
+        const messages = entries.slice(-maxCount)
+
+        if (select) {
+            const sourceIds = [...new Set(messages.map(m => m.source?.id).filter(id => id && this.RED.nodes.node(id)))]
+            if (sourceIds.length > 0) {
+                this.selectNodes(sourceIds)
```
Steve-Mcl: Why would getting debug logs select nodes?

Steve-Mcl (commented):
Part of the problem with this is the logs are fleeting on a fast moving instance. 

there was a lot of consideration went into the existing functionality (pinning logs of interest so as not to be lost as more unrelated log entries scroll in), also taking into account only visible logs (user could be filtering / viewing a different workspace tab to where the logs are being sent) 

anyhow, why is there an option to select the nodes? and no option to pin the debug messages?

I think this needs more careful thought and design.

resources/expertAutomations.js
andypalmi: Screenshot 2026-08-24 at 17 28 15

It's to replicate the UI functionality of selecting a specific set of nodes in the logs. Logs are currently limited to 100 IIRC with FIFO logic (first in first out) but this would basically allow the agent to only get the debug logs from an array of nodes

andypalmi:
Thanks for the pushback, this is worth getting right.

On `select`: fair point, dropping it. It was meant to let the agent highlight the nodes behind a set of messages, but there's already a standalone select-nodes action the agent can call with the same node ids from the results, so it was just duplicating an existing capability for no reason. Removed it from the schema, the automation, and the flow-builder tool.

On filtering by tab/workspace: the debug sidebar's own tab/node filter is UI-side, not something the comms channel enforces. Every debug message still gets delivered to the client and appended to the sidebar's list regardless of what tab or node filter the user currently has selected; the filter just adds a `hide` CSS class to the entries that don't match. `getDebugMessages` reads that same list with `visibleOnly` defaulted to false, so it sees everything in the list regardless of the user's current filter/tab, same as it would if nothing were filtered.

On pinning: agreed the underlying limitation is real; retention is capped by Node-RED's own list (FIFO, same cap as the sidebar), so a message can fall off before the agent ever asks for it, same as it would for a human who didn't pin it. We're leaving that as-is for now. Once a message is fetched it lands in the agent's chat history, so it's preserved there regardless of what happens in the sidebar afterward, and we'd rather see if that's a real problem in practice before adding pin-awareness. Happy to revisit if it turns out to bite.

Steve-Mcl:
> Thanks for the pushback, this is worth getting right.
> 
> On `select`: fair point, dropping it. It was meant to let the agent highlight the nodes behind a set of messages, but there's already a standalone select-nodes action the agent can call with the same node ids from the results, so it was just duplicating an existing capability for no reason. Removed it from the schema, the automation, and the flow-builder tool.
> 
> On filtering by tab/workspace: the debug sidebar's own tab/node filter is UI-side, not something the comms channel enforces. Every debug message still gets delivered to the client and appended to the sidebar's list regardless of what tab or node filter the user currently has selected; the filter just adds a `hide` CSS class to the entries that don't match. `getDebugMessages` reads that same list with `visibleOnly` defaulted to false, so it sees everything in the list regardless of the user's current filter/tab, same as it would if nothing were filtered.
> 
> On pinning: agreed the underlying limitation is real; retention is capped by Node-RED's own list (FIFO, same cap as the sidebar), so a message can fall off before the agent ever asks for it, same as it would for a human who didn't pin it. We're leaving that as-is for now. Once a message is fetched it lands in the agent's chat history, so it's preserved there regardless of what happens in the sidebar afterward, and we'd rather see if that's a real problem in practice before adding pin-awareness. Happy to revisit if it turns out to bite.

You sound like a bot ;)

Steve-Mcl (approved):

PR #395, opened 2026-08-25
feat(automations): apply set-wires as a batch with partial success

andypalmi:
## Summary

The `automation/set-wires` action now accepts a batch of wires in a single call and applies each connection independently, instead of connecting one wire per call. This removes the round-trip-per-wire cost when wiring up a flow.

## What changed

The action payload gains a `wires` array, each entry being `{ source, output?, target }`, with a single top-level `mode` (`add` or `remove`) shared by every wire in the batch. The single-wire payload shape is still accepted, so existing callers keep working.

Each wire is applied on its own: a wire that fails validation is recorded rather than aborting the batch. The result reports `{ mode, applied, failed }`, where `failed` entries carry the wire summary plus the error message. Write permission is checked once up front, so a permission failure still rejects the whole call.

Internally the per-wire validation and mutation moved into an apply that throws on failure. The batch loop turns those throws into `failed` entries, and the subroutine builder reuses the same apply so its rollback-on-throw behaviour is unchanged. The canvas is repainted once after the batch rather than per wire.

## Testing

`npm test` (475 passing). Added coverage for batches, mixed valid/invalid batches, the legacy single-wire shape, empty input, and the up-front permission check. Validated end to end against a running instance: a five-wire add applied all five with an empty `failed` list.

### DIFF
--- a/resources/expertAutomations.js
+++ b/resources/expertAutomations.js
@@ -367,12 +367,22 @@ export class ExpertAutomations extends ExpertActionsInterface {
             params: {
                 type: 'object',
                 properties: {
-                    mode: { type: 'string', enum: ['add', 'remove'], description: 'Whether to add or remove a wire' },
-                    source: { type: 'string', description: 'Source node ID' },
-                    output: { type: 'number', description: 'Source output port index (0-based)' },
-                    target: { type: 'string', description: 'Target node ID' }
+                    mode: { type: 'string', enum: ['add', 'remove'], description: 'Whether to add or remove every wire in the batch' },
+                    wires: {
+                        type: 'array',
+                        description: 'The wires to add or remove, applied in order',
+                        items: {
+                            type: 'object',
+                            properties: {
+                                source: { type: 'string', description: 'Source node ID' },
+                                output: { type: 'number', description: 'Source output port index (0-based, defaults to 0)' },
+                                target: { type: 'string', description: 'Target node ID' }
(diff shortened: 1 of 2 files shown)


Steve-Mcl (commented):
not a full review (its late) but should we leave the old SET_WIRES as-is and create a SET_WIRES_MULTIPLE. that would be less intrusive and easier to integrate in mcp tooling!

andypalmi:
@Steve-Mcl I went with `v1`/`v2` to match `manage_groups_v1/v2` where also the action shape changed between versions. 

It's not really less work as a separate tool: the batch handler only exists in 0.19.0+, so `set_wires_multiple` would still need the same version gate. 

This way the model just sees one `set_wires` at a time. Happy to switch if you'd prefer though.

andypalmi:
Also with a single `set_wires` we'd have to pass the `nr-assistant` version to the flow-building server so it knows which shape to send, which we don't. 

Version-gating v1/v2 lets us control which tool is available per version instead like we already do for managing groups.

Does this sound right to you?

@Steve-Mcl

Steve-Mcl (approved):
ok, but I think we should fast fail and not do partial edits.

for example, we should scan the source and target ids are present and on editable tabs BEFORE applying any wire changes so as not to make partial edits?

WDYT? You decide, merge if you are ok with current implementation. It could be a follow up?

